PR #8, opened 2024-01-26
Add JVS INGT10 GQS60EP T10A+OV9750

johndi3:


FlyRouter (commented):
In my opinion, these actions are unnecessary
We need to make sure that this driver is compiled from source directly into Firmware along with the rest.

```
devices/t10_lite_jvs-ingt10-gqs60ep/general/package/ingenic-opensdk/0001-add-ov9750.patch
devices/t10_lite_jvs-ingt10-gqs60ep/general/package/ingenic-osdrv-t20/ingenic-osdrv-t20.mk
```

cronyx:
> In my opinion, these actions are unnecessary We need to make sure that this driver is compiled from source directly into Firmware along with the rest.
> 
> ```
> devices/t10_lite_jvs-ingt10-gqs60ep/general/package/ingenic-opensdk/0001-add-ov9750.patch
> devices/t10_lite_jvs-ingt10-gqs60ep/general/package/ingenic-osdrv-t20/ingenic-osdrv-t20.mk
> ```

At the moment there is no other solution, you need to either make a patch adding a sensor to openingenic or add an already compiled driver to the assembly.

FlyRouter:
Okay, then this is the proposal.
We accept this patch, but at the same time open a ticket in the openingenic repo about the need to create a patch for this driver.
Once created and verified in openingenic, I think we can remove the files from the builder repo.
I mean that you need to try to do as much as possible in other repositories so that the builder repo remains minimalistic for each device.
Thanks for understanding.

FlyRouter (approved):

FlyRouter:
You need to enable compilation of this sensor in the openingenic repository and, after checking, remove the files from the device to the ingenic repo. We need to strive for universality, so that as few change files as possible are in the Builder repo.
Only a package configuration file, an unnecessary removal file, a settings file. Three files are the minimalism you need to strive for

PR #11, opened 2024-02-05
add h3c tc2101

jimsmt:


FlyRouter (approved):
Thanks for the PR of the new device
Are you sure there is no control over the GPIO WiFi module?

Please add a QR code reading package, as done in others.
BR2_PACKAGE_QUIRC_OPENIPC=y
We are adding it to all WiFi devices in the Builder repository.
..._lite_h3c-tc2101/br-ext-chip-sigmastar/configs/ssc337_lite_h3c-tc2101_defconfig

Please make a note for the future that we need to optimize the motor code for SigmaStar so that options can be specified as arguments or profiles
devices/ssc337_lite_h3c-tc2101/general/package/sigmastar-motors/src/motor.c

FlyRouter:
Please be sure to specify GPIO for WiFi management in the configuration file even if it works by default. In the future there will be a daemon that will monitor the connection and if there is no connection with the access point, it will try to turn off/on the modem. so this data is important. Thanks for understanding.

PR #28, opened 2024-08-02
fix memory config for xiaomi-mjsxj03hl-jxq03

kiryam:


flyrouter (commented):
@cronyx 
What do you think about these corrections?

PR #29, opened 2024-08-03
Update customizer.sh

djcfg123:


flyrouter (commented):
Why were the commented settings edited?

PR #34, opened 2024-11-25
adding devices/ssc333_lite_tp-link-tapo-c110-v26

geedubess:
Depends on https://github.com/OpenIPC/firmware/pull/1619.

I'm not sure if ATBM_WIFI is optimal (it is probably not); this is probably related to naming in the package name (above pull request).

This wifi chip has been spotted in v2.2 boards TAPO C110 boards as well, so it could be named something different, but that suggests the existing board name "ssc333_lite_tp-link-tapo-c110-v2" should also be renamed (eg ssc333_6012b_lite_tp-link-tapo-c110).


flyrouter (approved):
Thank you

PR #58, opened 2025-06-08
Add Jienuo JN-107AR-E-WIFI

davidebeatrici:
SD card slot untested, everything else is confirmed working.

U-Boot password: `HI2105CHIP`

https://github.com/thegroove/openipc-wiki/blob/device-jienuo-jn-107ar-e-wifi/en%2Fdevice-jienuo-jn-107ar-e-wifi.md

<details>
<summary>Pictures</summary>

</details>

<details>
<summary>Possible successor (JN-603-D-WIFI)</summary>

</details>

@thegroove


davidebeatrici:
Unfortunately there seems to be no way to control the IR LEDs, which are always on.

flyrouter (commented):
Good day
You have done a good job, but it is highly recommended to switch the firmware to Lite mode and not Ultimate using a script to exclude unnecessary components as in other cameras.

openipc-ai:
Thanks for this, and apologies for the long silence. Reviewed against current master — it is the tidiest of the device PRs currently open, and it should land.

**Good news:** you no longer need a `master.yml` matrix entry. Since #120 the build matrix is read off the tree, so a device is registered simply by having `devices/<dir>/.../<dir>_defconfig`. This repo also builds pull requests now, narrowed to the devices a diff can reach — so a rebase push here will build `t31_ultimate_jienuo-jn-107ar-e-wifi` for the first time, and only that device.

**What checks out**

- `fw_setenv wlandev rtl8733bu-t31-camhipro` exists upstream in firmware's `general/overlay/etc/wireless/usb`.
- Directory name, defconfig filename and the `-nor.tgz` in the `upgrade` URL all agree, which is what keeps self-update working.
- No excludes list is fine here: it's a 16M target, and neither `t31_ultimate_*` device in the tree ships one.

**Two things before merge**

1. The branch now conflicts with master — needs a rebase.
2. While rebasing, drop `BR2_PACKAGE_JSONFILTER=y`. #128 removed it from all 97 defconfigs after firmware/#2304; it drags in libubox for nothing on a device that enables neither uqmi-openipc nor wifibroadcast-ng, and it had three boards over their rootfs cap.

On the IR LEDs being stuck on — that doesn't block the merge, but would you mind reflecting it in the README row rather than `done`? Other rows use that form, e.g. `video-yes, wifi-yes, motors-no`. It saves the next person with this camera from assuming it's solved.

openipc-ai (changes requested):
Marking this changes-requested so it shows up as such in the queue. Nothing new — the same two items from my review above are still outstanding:

1. The branch conflicts with master and needs a rebase.
2. While rebasing, drop `BR2_PACKAGE_JSONFILTER=y` (#128 removed it from all 97 defconfigs after firmware/#2304).

The README status wording for the always-on IR LEDs is still a nice-to-have, not a blocker.

openipc-ai:
Picking this back up rather than letting it age out. The device profile itself
was fine — only the rebase and the `jsonfilter` line were outstanding — so I
have done both and pushed to your branch.

A note on how, because the history changed more than a plain rebase would:

The branch carried a third commit, `a44bc34` "Update ssc338q_apfpv_defconfig",
by another contributor, which edited
`devices/ssc338q_apfpv/general/overlay/usr/share/openipc/customizer.sh`. That
file no longer exists in master, so replaying it would only have produced a
conflict over an unrelated device. Your net change never touched apfpv anyway.
The README hunk had also gone stale against the current device table.

So rather than fight the replay I rebuilt the branch as your net change on top
of master: the two device files exactly as you wrote them, plus the README row
reinserted in its correct alphabetical slot, with `BR2_PACKAGE_JSONFILTER=y`
dropped. Authorship on the commit is still yours.

This was a force-push, so reset to the remote rather than merging if you have
the branch locally.

`ci-matrix --self-test` and the `cli` path lint pass, and the matrix narrows to
`t31_ultimate_jienuo-jn-107ar-e-wifi`. Build approved; if it is green I will
merge. Apologies this sat for so long.

davidebeatrici:
Thank you!

PR #82, opened 2026-03-26
Add Smitch 360 PTZ (HI3518EV300, JXF23, RTL8188FU, NOR 16MB)

siddhant01:
## Device details

- **Model**: Smitch 360 PTZ (Indian brand, servers closed)
- **Board**: T5810hct-m02
- **SOC**: HI3518EV300
- **Flash**: EN25QH128 (16MB NOR)
- **RAM**: 64MB
- **WiFi**: RTL8188FU (internal USB, GPIO 40 power-up)
- **Sensor**: JXF23
- **Variant**: ultimate

## Notes

- WiFi power-up via GPIO 40 — wlandev profile `rtl8188fu-hi3518ev300-t5810hct-m02`
  already merged in firmware repo: OpenIPC/firmware#1373
- QR WiFi provisioning tested and working
- Motors, IR cut, LEDs, reset button not yet mapped (GPIO unknown)

## Status

video-yes, wifi-yes, motors-no


openipc-ai:
Thanks for this, and sorry it sat so long. It is in good shape — reviewed against current master.

**Good news first:** you no longer need a `master.yml` matrix entry. Since #120 the build matrix is read off the tree, so a device is registered simply by having `devices/<dir>/.../<dir>_defconfig`. Your device therefore registers itself, and this repo now builds pull requests — a push here will build `hi3518ev300_ultimate_smitch-360ptz-t5810hct-m02` for the first time, narrowed to just that one device. (CLAUDE.md step 6 still describes the old flow; that's on us to fix.)

**What checks out**

- `fw_setenv wlandev rtl8188fu-hi3518ev300-t5810hct-m02` — already upstream in firmware's `general/overlay/etc/wireless/usb`, complete with a `# HI3518EV300 t5810hct-m02` comment. So the Wi-Fi half is landed.
- The exclude list correctly *keeps* `libsns_f23.so`, `jxf23_i2c_1080p.ini` and `iq/f23.ini` for a JXF23 board, and its name (`hi3518ev300_ultimate.list`) matches the soc+flavour convention.
- The UBI block matches what both `hi3516ev300_ultimate_*` devices do, so that's right for the family.

**Three small things before merge**

1. Drop `BR2_PACKAGE_JSONFILTER=y`. #128 removed it from all 97 defconfigs after firmware/#2304, because it drags in libubox for nothing on a device that enables neither uqmi-openipc nor wifibroadcast-ng — it was worth 32KB and had three boards over their rootfs cap.
2. Please drop the README wording change on the "Create firmware with built-in credentials" heading — it's unrelated to the device and just makes the diff harder to read. The device-table row is welcome, keep that.
3. `customizer.sh` is missing its trailing newline.

Rebase on master with those and CI will tell us whether it builds.

openipc-ai (changes requested):
Marking this changes-requested so it shows up as such in the queue. Nothing new — the same three items from my review above are still outstanding:

1. Drop `BR2_PACKAGE_JSONFILTER=y` from the defconfig (#128 removed it from all 97 defconfigs after firmware/#2304).
2. Drop the unrelated README wording change on the "Create firmware with built-in credentials" heading. The device-table row is welcome, keep that.
3. `customizer.sh` is missing its trailing newline.

The branch still merges cleanly, so no rebase is needed — push those three and CI will build the device.

openipc-ai:
Picking this back up rather than letting it age out — the device work was fine,
only the three small items were outstanding, so I have pushed them to your
branch myself:

- dropped `BR2_PACKAGE_JSONFILTER=y` from the defconfig;
- restored the `repack.sh` section heading to what master has (your device-table
  row is untouched);
- added the missing trailing newline to `customizer.sh`.

I also rebased onto current master. When I last looked the branch still merged
cleanly, but the README device table has taken a number of new rows since, and
your row now collided with the SpezVision SVI-252B entry. Both are kept, with
Smitch sorted ahead of SpezVision.

Note this was a force-push, so if you have a local copy of the branch, reset to
the remote rather than merging.

`ci-matrix --self-test` and the `cli` path lint both pass, and the matrix
narrows to `hi3518ev300_ultimate_smitch-360ptz-t5810hct-m02`. I have approved
the build; assuming it comes back green I will merge. Thanks for the profile,
and sorry it sat as long as it did.

openipc-ai:
Merged. One more fix was needed beyond the three above, worth writing down
because it is a real property of the profile rather than paperwork.

The first build failed late, after generating `rootfs.ubifs`:

    sed: can't read .../general/scripts/ubifs/ubinize_hisilicon.cfg: No such file or directory
    make[1]: *** [fs/ubi/ubi.mk:51: .../rootfs.ubi] Error 2

`BR2_TARGET_ROOTFS_UBI_CUSTOM_CONFIG_FILE` pointed at `ubinize_hisilicon.cfg`,
which is not in the firmware tree — only `ubinize.cfg`, `ubinize_rockchip.cfg`
and `ubinize_sigmastar.cfg` are. The two devices that use that name each ship
their own private copy under `general/scripts/ubifs/` in their device
directory, and this profile referenced the name without carrying the file.

Worth knowing that the UBI block is not dead weight on a NOR board, so deleting
it would have been the wrong fix: CI publishes both a `-nor.tgz` and a
`-nand.tgz` for these `ultimate` profiles, and the nand repack consumes
`rootfs.ubi`. Dropping it would have silently stopped producing the NAND image.

I switched it to the in-tree `ubinize.cfg`, which is what both hi3518ev200
`ultimate` boards use. That suits the nand repack better than the hisilicon
variant, which adds a `[kernel]` volume while the repack already passes
`uImage` separately. Your NOR image is unaffected — that path repacks `uImage`
plus `rootfs.squashfs`.

Both now build with room to spare: NOR rootfs 6828/8192 KB, NAND rootfs.ubi
13568/16384 KB, kernel 1873/2048 KB.

If the motors get working, a follow-up updating the README status would be
welcome. Thanks for the profile, and again, sorry it waited so long.

PR #84, opened 2026-04-11
Support transition from hi3516ev200_isp to open_isp @ hi3516ev200

PeterBeklemishev:
Previusly discussed in https://github.com/OpenIPC/firmware/issues/1991


flyrouter (approved):
Thank you

PR #97, opened 2026-05-26
Williampunk

Williampunk:


openipc-ai (changes requested):
Thanks for working on the Qihoo D706 — a T20 board with no usable Ethernet PHY is a useful addition. But this branch can't merge as it stands, because of how the new device was created.

### Blocking: this rewrites an existing, working camera

The D706 was added by editing `devices/t20_ultimate_azarton-c1/` in place instead of copying it to a new directory. The Azarton C1 (T20X, JXF23, `RTL8189FS_SDIO`, listed `done` in the README) is a separate shipping device, and on this branch it becomes a second copy of the D706:

- `fw_setenv upgrade '.../t20_ultimate_azarton-c1-nor.tgz'` is **deleted** — the camera loses self-update entirely.
- `BR2_OPENIPC_SNS_MODEL` changes `jxf23` → `gc2023`, and the customizer starts setting `fw_setenv sensor gc2023`. The Azarton has a JXF23.
- `fw_setenv osmem` drops 64M → 40M and `rmem` 64M → 24M.
- `BR2_OPENIPC_VARIANT` flips `ultimate` → `lite` while the directory is still named `t20_ultimate_...`, and majestic-webui, fonts, opus, lame, quirc, mosquitto and wireguard are dropped from it.
- `fw_setenv extras nogmac` is added, which is a D706 property — the Azarton has working Ethernet.

Anyone flashing the Azarton nightly after this would get a broken image. It is also a `SMOKE_TARGET` in `.github/scripts/ci-matrix.py` (it covers Ingenic `ultimate`), and I confirmed `--self-test` still passes with this branch applied — so CI would not catch it either.

**Fix:** revert everything under `devices/t20_ultimate_azarton-c1/` back to master and keep only the new `devices/t20_lite_qihoo-*/` directory. Copying a sibling and editing the copy is the intended flow.

### Also needed

1. **Directory case.** The directory is `t20_lite_qihoo-D706` but the defconfig inside is `t20_lite_qihoo-d706_defconfig`. Device directory names are lowercase and must match the defconfig basename — that name *is* the `BOARD` token. Please rename the directory to `t20_lite_qihoo-d706`.
2. **No `upgrade` URL.** The new `customizer.sh` never calls `fw_setenv upgrade`, so the device can't self-update. It needs `fw_setenv upgrade 'https://github.com/OpenIPC/builder/releases/download/latest/t20_lite_qihoo-d706-nor.tgz'`, with the filename matching the directory name exactly.
3. **Drop `BR2_PACKAGE_JSONFILTER=y`.** #128 removed it from all 97 defconfigs after firmware/#2304 — it drags in libubox for nothing unless the device enables uqmi-openipc or wifibroadcast-ng.
4. **PR title.** "Williampunk" doesn't say what the change does; something like `t20: add Qihoo D706 (GC2023, RTL8189FS)` would help.

No `master.yml` entry is needed any more, by the way — since #120 the matrix is read off the tree, so the new directory registers itself and PR CI will build it, narrowed to just that device.

openipc-ai:
Closing this for queue hygiene — it has been open since May with the blocking
item unaddressed. That is not a judgement on the Qihoo D706 work, which is
still welcome.

To recap what would need to happen, so nothing is lost if you come back to it:
the branch edits `devices/t20_ultimate_azarton-c1/` in place rather than copying
it, so on this branch the Azarton C1 becomes a second copy of the D706 — it
loses its `upgrade` URL, its sensor changes `jxf23` to `gc2023`, its osmem/rmem
drop, and its variant flips to `lite` while the directory is still named
`ultimate`. Anyone flashing the Azarton nightly would get a broken image. The
fix is to revert everything under that directory back to master and keep only
the new `devices/t20_lite_qihoo-d706/` directory.

Alongside that: lowercase the new directory so it matches its defconfig
basename, add the `fw_setenv upgrade` line pointing at
`t20_lite_qihoo-d706-nor.tgz`, and drop `BR2_PACKAGE_JSONFILTER=y`.

If you want to pick it up again, please open a fresh PR with the Azarton
directory untouched and I will review it promptly. Thanks for the contribution,
and sorry it stalled.

PR #110, opened 2026-07-05
package: general waybeam encoder package for Infinity6E and Infinity6C

snokvist:
Adds a Buildroot package `waybeam`: an H.265 video encoder and streamer for
SigmaStar Infinity6E (SSC338Q/SSC30KQ) and Infinity6C (SSC378QE). The SoC
backend is selected from `OPENIPC_SOC_FAMILY`. The source is MIT-licensed and
tracks `OpenIPC/waybeam_venc` master (currently v0.40.1).

Installs:
- `/usr/bin/waybeam`
- `/usr/bin/json_cli` (config editor)
- `/etc/waybeam.json`
- `/etc/init.d/S95waybeam`

**Unlocked sensor modules (both families).** Waybeam ships its own imx335/imx415
drivers with higher-FPS sensor modes:
- **Infinity6E (Star6E):** installs the unlocked `sensor_imx{335,415}_star6e.ko`
  over the stock `/lib/modules/4.9.84/sigmastar/sensor_imx*_mipi.ko` names
  (up to 144 fps IMX335 / 100 fps IMX415). No custom ISP bins — the SDK/stock
  tuning applies (`isp.sensorBin` defaults to empty).
- **Infinity6C (Maruko):** installs `sensor_imx{335,415}_mipi.ko` plus the
  matching `/etc/sensors/imx335.bin` and `imx415.bin`, since there are no
  in-tree modules for these sensors on I6C.

Both are ordered after `sigmastar-osdrv-sensors` and the family's
`sigmastar-osdrv-infinity6{e,c}` so the modules install over the stock ones.

waybeam and Majestic cannot be built together (`depends on
!BR2_PACKAGE_MAJESTIC`): both drive the same sensor and encoder, and the
SigmaStar OSDRV MI libraries are installed to `/usr/lib` only when Majestic
is not selected. The package ships no libraries; OSDRV provides them.

Defaults in `/etc/waybeam.json`: UDP RTP output enabled to
`udp://192.168.1.10:5600`, resolution `auto` (sensor native), ROI and audio
disabled. `isp.sensorBin` is empty, which is valid — the encoder falls back
to the SDK default ISP tuning, so video comes up on first boot. The config is
kept in sync with the v0.40.1 schema (carries the `attitude` section; the dead
`frameLost` field, removed from the encoder in v0.19.0, was dropped).

Verified with scoped Buildroot builds on both families (Majestic off). Each
uses the Buildroot toolchain; the Infinity6C build installs the binary,
json_cli, both sensor modules, both ISP bins, config and init script:
- Infinity6E — glibc binary.
- Infinity6C (SSC378QE) — musl binary.

**Verification note for the Star6E sensor modules:** they ship prebuilt
(stripped `--strip-debug`) and must match the target kernel's vermagic. They
were built against the OpenIPC Infinity6E 4.9.84 kernel; a scoped I6E build +
boot check that `sensor_imx335_mipi.ko` / `imx415.ko` `insmod` cleanly and the
new modes enumerate is the remaining gate (the Maruko modules were validated
the same way).

### DIFF
--- a/package/waybeam/Config.in
+++ b/package/waybeam/Config.in
@@ -1,0 +1,22 @@
+config BR2_PACKAGE_WAYBEAM
+	bool "waybeam"
+	depends on BR2_TOOLCHAIN_HAS_THREADS
+	depends on BR2_PACKAGE_SIGMASTAR_OSDRV_INFINITY6E || \
+		BR2_PACKAGE_SIGMASTAR_OSDRV_INFINITY6C
+	depends on !BR2_PACKAGE_MAJESTIC
+	help
+	  Standalone H.265 video encoder and streamer for SigmaStar
+	  Infinity6E (SSC338Q/SSC30KQ) and Infinity6C (SSC378QE). The
+	  SoC backend is selected automatically from the target SoC
+	  family. Provides a dual encoder backend, sensor unlock and an
+	  HTTP configuration API.
+
+	  It is an alternative to the Majestic streamer and cannot be
+	  built alongside it: both drive the same sensor and encoder
+	  hardware, and both rely on the SigmaStar OSDRV MI libraries,
+	  which OSDRV installs only when Majestic is not selected.
+
+	  Installs:
+	    - /usr/bin/waybeam
+	    - /etc/waybeam.json
+	    - /etc/init.d/S95waybeam
--- a/package/waybeam/files/S95waybeam
+++ b/package/waybeam/files/S95waybeam
@@ -1,0 +1,60 @@
+#!/bin/sh
+
+DAEMON="waybeam"
+RESPAWN="waybeam-resp"
+WATCHDOG="waybeam-wd"
+DAEMON_PATH="/usr/bin/waybeam"
+LOG_PATH="/tmp/waybeam.log"
+
+# waybeam forks a respawn child (comm "waybeam-resp") on an API restart and a
+# watchdog ("waybeam-wd"). pidof and killall match by comm, so every restart
+# path must cover all three names, otherwise stop() returns while a helper is
+# still alive and the next start() trips the duplicate-instance check.
+still_running() {
+	pidof "$DAEMON" "$RESPAWN" "$WATCHDOG" >/dev/null 2>&1
+}
+
+wait_exit() {
+	i=0
+	while [ $i -lt 30 ]; do
+		still_running || return 0
+		sleep 0.5
+		i=$((i + 1))
+	done
+	echo "Warning: $DAEMON did not exit in time"
+}
+
+start() {
+	if still_running; then
+		echo "$DAEMON already running"
+		return 0
+	fi
+	echo -n "Starting $DAEMON: "
+	# waybeam reads /etc/waybeam.json from a fixed path; it has no -c flag.
+	"$DAEMON_PATH" > "$LOG_PATH" 2>&1 &
+	i=0
+	while [ $i -lt 10 ]; do
+		sleep 0.3
+		still_running && { echo "OK"; return 0; }
+		i=$((i + 1))
+	done
+	echo "FAILED (crashed at init or lost the duplicate-instance check, see $LOG_PATH)"
+	return 1
+}
+
+stop() {
+	echo -n "Stopping $DAEMON: "
+	killall "$DAEMON" "$RESPAWN" "$WATCHDOG" 2>/dev/null
+	wait_exit
+	echo "OK"
+}
+
+case "$1" in
+	start)   start ;;
+	stop)    stop ;;
+	restart) stop; start ;;
+	*)
+		echo "Usage: $0 {start|stop|restart}"
+		exit 1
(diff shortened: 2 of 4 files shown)


flyrouter:
Good afternoon. 
Thanks for sending the PR. 
There are almost no issues with its formatting. 
However, in the next few days we will need to clarify a few organisational matters and I will let you know in PM before merging. 
Thanks.

snokvist:
I am happy to move this to OpenIPC/firmware PR instead if thats a more suitable place for it. At least one other uses has confirmed this package works for building waybeam.

openipc-ai:
**This has moved to OpenIPC/firmware — draft PR: OpenIPC/firmware#2332**

@snokvist, the package serves whole SoC families rather than one retail model,
so it belongs in the shared tree rather than in builder. Your commit is carried
over unsquashed with you as its author. It's a draft, and your review is what it
is waiting on — I'd have added you through the reviewer field, but GitHub only
accepts review requests from repository collaborators, so this is me asking
directly.

Three things changed on the way over, all to clear gates firmware enforces and
builder doesn't. The first two are worth your attention because they change what
your package does:

**The sensor modules are now compiled from source.** The prebuilt
`sensors/star6e/sensor_imx*_star6e.ko` are a binary without a buildable source in
firmware's terms, and they have to be vermagic-matched by hand — which is exactly
why your description lists an I6E `insmod` check as the remaining gate. Since
`drivers/sensor_imx*_{star6e,maruko}.c` are real sources, the package now builds
them against the kernel from the same build:

```make
$(MAKE) -C $(@D)/drivers sensor SOC=$(WAYBEAM_SOC) \
        KSRC="$(LINUX_DIR)" CROSS="$(TARGET_CROSS)"
```

A local `ssc338q_lite` build (Majestic off) confirms they compile and land with
the right vermagic: `vermagic=4.9.84 SMP preempt mod_unload ARMv7 thumb2 p2v8`.
That should close your open gate structurally rather than by testing. This needs
`WAYBEAM_GIT_SUBMODULES = YES`, since the SigmaStar headers come from the
`sensors-src/` submodule on OpenIPC/sensors, and the install path follows
`LINUX_VERSION_PROBED` instead of a literal `4.9.84` / `5.10.61`.

**The Infinity6C ISP tuning bins are not installed.**
`iq-profiles/maruko-bin/README` records them as pulled from a working SSC378QE
bench `/etc/sensors/`, and firmware refuses binaries lifted off a running camera.
It should cost nothing: `waybeam.json` ships `isp.sensorBin` empty, and
`sigmastar-osdrv-sensors` already puts `imx335.bin` and `imx415.bin` in
`/etc/sensors/`. **If your Maruko bench depends on those specific bins rather
than the stock ones, say so** — that's the change most likely to be wrong, and
you're the only one who can tell.

Third and minor: the site follows the `OpenIPC/waybeam_venc` → `OpenIPC/waybeam`
rename. `WAYBEAM_VERSION` stays `HEAD`, matching the other OpenIPC-org packages
in the tree.

What it still needs is hardware, which is why it's a draft: an SSC338Q or SSC30KQ
and an SSC378QE booting an image with it, confirming the modules `insmod`
cleanly, the unlocked modes enumerate, and the stream comes up with
`isp.sensorBin` empty. If you can run that and paste the output on #2332, it can
come out of draft.

openipc-ai:
Closing this as superseded: the package landed in the shared tree on 29 August as OpenIPC/firmware#2332, squashed to `47ccdbee` with you as co-author. `general/package/waybeam` is in firmware master now, so this PR duplicates shipped code.

Thanks for the review on the migration — it changed the result in three ways worth recording here for anyone who finds this PR later:

- **Sensor modules are compiled from source.** `drivers/*.c` built against the kernel of the same build, `WAYBEAM_GIT_SUBMODULES = YES` for the SigmaStar headers from the OpenIPC/sensors submodule, and `LINUX_VERSION_PROBED` for the install path — so vermagic cannot drift, which closes the I6E `insmod` gate this PR left open.
- **The vendored Infinity6C ISP bins are dropped**, being byte-identical to `sigmastar-osdrv-infinity6c`'s own copies. With `isp.sensorBin` empty, waybeam resolves `/etc/sensors/<sensor>.bin` itself — the same stock tuning Majestic loads.
- **Infinity6C is not offered upstream.** Your SSC378QE bench found the encoder running against MI libraries it was not built for: waybeam's I6C backend targets the 2024-06-18 SigmaStar MI set, `sigmastar-osdrv-infinity6c` installs 2022-09-07. It streams, but AE never actuates — mean luma 2.05/255 against 153.91 with the 2024 libraries. I6C comes back once those revisions are reconciled in the OSDRV package, where a library bump affects every I6C board.

One practical reason not to leave this one open: `copy_extra_packages` in `builder.sh` does `cp -afv package/* openipc/general/package/`, so a builder-local `package/waybeam` would now silently overwrite the upstream one on every build, and the only symptom would be the older package being built.

PR #111, opened 2026-07-11
ci: add t31_lite_tp-link-tapo-tc70-v3

NHellFire:
Realised I'd forgotten to do this.

### DIFF
--- a/.github/workflows/master.yml
+++ b/.github/workflows/master.yml
@@ -134,6 +134,7 @@ jobs:
           - t31_lite_tuya-gv7630-t31-ptz
           - t31_ultimate_azarton-c1-t31x
           - t31_ultimate_gcraftsman-gca50
+          - t31_lite_tp-link-tapo-tc70-v3
           - t31_lite_zte-k540
           - t40_lite_movols-mo-805p
 


openipc-ai:
Thanks for this, and sorry it went unanswered. The device you're asking for is the right ask — `t31_lite_tp-link-tapo-tc70-v3` has a complete `devices/` directory and has never once been built.

The mechanism moved out from under this PR, though. Since #120 the build matrix is no longer a list in `master.yml`; it is derived from the tree, and the devices that existed but were in no matrix at the time were frozen into a `NOT_BUILT` opt-out in `.github/scripts/ci-matrix.py` so that a device falling out of CI would be a test failure rather than a silent skip. The TC70 v3 is one of those seven, which is why this branch now conflicts — the block it appends to is gone.

I've opened #132, which removes it (and the KC110, from #108) from `NOT_BUILT`. That's this PR's request, expressed the way the selector now wants it. Since a change to `ci-matrix.py` cannot narrow the matrix, #132 runs the full 109-device build, so the TC70 v3 gets proven in that run.

Leaving this open until #132 lands, then closing it as done rather than rejected.

openipc-ai:
Done — closing this as **completed**, not rejected. What you asked for is on master.

#132 landed as 9df259c and removed `t31_lite_tp-link-tapo-tc70-v3` from the `NOT_BUILT` opt-out in `.github/scripts/ci-matrix.py`, which is where device registration moved when #120 replaced the `master.yml` matrix with one derived from the tree. That is why this branch went stale through no fault of yours: the block you appended to no longer exists, and CLAUDE.md was still telling contributors to edit it (fixed separately in #134).

On master now:

```
$ echo devices/t31_lite_tp-link-tapo-tc70-v3/br-ext-chip-ingenic/configs/\
t31_lite_tp-link-tapo-tc70-v3_defconfig | python3 .github/scripts/ci-matrix.py --stdin
ci-matrix: 1/110 devices (needs_build=True) --- narrowed to the affected devices
  t31_lite_tp-link-tapo-tc70-v3
```

It also went through its first-ever build in #132's run and came out green, so the device is genuinely building, not just listed. It is in the nightly from here on, and any future change under `devices/t31_lite_tp-link-tapo-tc70-v3/` will build just that device on the PR.

Thanks for flagging it — the device had been sitting in the tree unbuilt for months and this is what got it noticed.

PR #122, opened 2026-08-18
ci: close the if that cost last night's release, and parse-check the rest

widgetii:
## What broke

The Collect guard added to the publish job in #121 shipped without its closing `fi`.

[Run 32096242353](https://github.com/OpenIPC/builder/actions/runs/32096242353) — the first nightly to reach the new single publish job — built all 107 devices, staged 219 assets, and then died in `bash -e`:

```
line 25: syntax error: unexpected end of file from `if' command on line 14
```

Nothing was written. There is no `nightly-20260818-*` release; `nightly` and `latest` still carry the 17th's images (assets last touched `2026-08-17T04:19`). `manifest.yml` ran afterwards and re-indexed yesterday's data. ci-gate went red — the gate did its job, there was simply nothing left to publish by the time it ran.

Without this, tonight's 03:00 cron fails the same way.

## Why CI was green

The publish job carries `if: github.event_name != 'pull_request'` — there are no artifacts to publish on a PR, so the whole path is unreachable from PR CI by construction. Both of #121's runs passed without ever executing the step.

#121 was explicit about this ("none of this path can run on a pull request") and checked the four guard states offline. But offline-checking the *logic* doesn't check that the script *parses*, and that's what broke. Any edit to that job ships unexecuted.

## What's here

1. The missing `fi` in `master.yml`.
2. `.github/scripts/lint-workflow-shell.py` — parses every `run:` block in `.github/workflows/` with `bash -n`.
3. `.github/workflows/lint.yml` — runs it on PRs and on master pushes touching either file. Seconds, no runners, no matrix.

### Notes on the checker

- **`${{ }}` is not shell**, so each expression is replaced with a plain word first. Real limitation: an expression interpolating shell syntax is checked as the word, not as what it expands to — the same limitation `bash -n` has with any variable. The substitution preserves line counts so reported lines still point at the right line of the workflow.
- **`--self-test` runs first in CI.** The way this class of checker breaks is by making *everything* pass, which is indistinguishable from a clean tree. So it asserts an unterminated `if` *containing an expression* is still rejected while the closed form passes.
- **Steps are found structurally** — any mapping with a `run:` key — not by the `jobs.*.steps[*]` path, so nothing depends on guessing where they live. A discovery floor fails the run if the walk stops finding them.
- **Syntax only, deliberately.** Says nothing about quoting or `-e` semantics. A shellcheck pass wants per-block disables for the substitution above and is a bigger, separate change.

## Verification

- Checker flags the merged `master.yml` with the same message the runner gave, pointing at `master.yml:354 (Collect assets)`
- Passes the fixed tree — 22 run blocks across all five workflows
- Self-test passes; exit codes correct (1 on broken, 0 on clean)
- All five guard states re-checked against the fixed block: normal night publishes; broken download and green-but-empty dist both fail with their intended errors; dead matrix with nothing built exits clean; partial matrix proceeds

## Related

`OpenIPC/firmware`'s backport (#2279) has the `fi` and is unaffected — its nightly ran clean at 22:46 UTC: 397 assets collected, paced uploads, dated release with 109 images + 96 sizes sidecars, both alias tags refreshed and moved to the built commit. Firmware has no equivalent workflow-shell lint either, so the same gap exists there; worth porting separately.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/.github/scripts/ci-matrix.py
+++ b/.github/scripts/ci-matrix.py
@@ -78,8 +78,8 @@ NOT_BUILT = {
 # whole filename, never as a prefix: one this list has never heard of is unknown,
 # and unknown widens. package.sh and repack.sh are developer and end-user tools
 # that no workflow invokes -- only builder.sh is on the CI path.
-NO_BUILD_WORKFLOWS = {"build-one.yml", "cleanup.yml", "manifest.yml"}
-NO_BUILD_SCRIPTS = {"enrich_manifest.py"}
+NO_BUILD_WORKFLOWS = {"build-one.yml", "cleanup.yml", "lint.yml", "manifest.yml"}
+NO_BUILD_SCRIPTS = {"enrich_manifest.py", "lint-workflow-shell.py"}
 NO_BUILD_FILES = {
     ".github/CODEOWNERS", ".gitignore", "LICENSE", "package.sh", "repack.sh",
 }
@@ -407,6 +407,8 @@ def self_test():
         ([".github/CODEOWNERS"], 0, "codeowners"),
         ([".github/workflows/manifest.yml"], 0, "manifest never builds"),
         ([".github/scripts/enrich_manifest.py"], 0, "manifest script"),
+        ([".github/workflows/lint.yml"], 0, "the workflow linter never builds"),
+        ([".github/scripts/lint-workflow-shell.py"], 0, "its script"),
         (["repack.sh"], 0, "end-user tool, no workflow runs it"),
         (["package.sh"], 0, "developer tool, no workflow runs it"),
         (["archive/gk7205v200_fpv/202607231714/openipc.tgz"], 0, "build output"),
--- a/.github/scripts/lint-workflow-shell.py
+++ b/.github/scripts/lint-workflow-shell.py
@@ -52,7 +52,14 @@ import sys
 import tempfile
 from pathlib import Path
 
-import yaml
+try:
+    import yaml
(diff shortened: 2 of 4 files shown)


widgetii:
Both findings addressed in 71eaed7.

**1. PyYAML not installed (High)** — real, and the fact that the job passed twice is the point: `ubuntu-latest` happens to ship PyYAML, so the linter had an undeclared dependency on whatever the runner image includes. Now installed explicitly with apt (24.04 is PEP 668, and it matches the busybox install in OpenIPC/firmware's `shell-tests`). The import is also guarded, so running it locally without PyYAML prints what to install instead of a traceback — verified by hiding the module on `PYTHONPATH`:

```
lint-workflow-shell: PyYAML is required and missing.
  CI installs it (see .github/workflows/lint.yml); locally:
    apt install python3-yaml   # or: pip install pyyaml
```

**2. Implicit token permissions (Medium)** — added `permissions: contents: read` at workflow level. The job reads the tree and reports; it writes nothing.

Neither would have failed the job today. Both fail loudly rather than quietly if they ever fire, which is the right direction, but an undeclared dependency and an inherited token are cheap now and annoying to diagnose later.

Re-verified after the change: self-test passes, full scan clean at 23 run blocks (the new apt step adds one), the linter still flags the merged `master.yml` at `master.yml:354 (Collect assets)` with the runner's own message, and `ci-matrix.py --self-test` passes at 33 cases.

PR #135, opened 2026-08-27
ci: let a pull request keep the image it just built

widgetii:
`Stage artifacts` and `Upload build artifacts` were gated on
`github.event_name != 'pull_request'`, sharing that guard with `Send binary` and
the `publish` job. Those two need it — one uses the Telegram bot secret, the
other writes shared release assets. The upload does not:
`actions/upload-artifact` is run-scoped and touches no secret and no shared
mutable state.

The effect is that a pull request compiles a complete firmware image and then
discards it. `Stage artifacts`, `Upload build artifacts`, `Send binary` and
`Publish releases` all report **skipped** on a green PR run, and the artifact
count is zero. A contributor adding a device profile therefore cannot obtain the
image their own PR just built — the only route to a flashable file is asking a
maintainer to dispatch `build-one` for them.

#119 hit exactly this. The author planned to "flash the CI artifact once this
branch produces one", which would never have resolved; I dispatched `build-one`
by hand to unblock it.

## Why this is safe

`publish` carries its own `github.event_name != 'pull_request'` in its `if:`,
independently of these steps:

```yaml
if: >-
  !cancelled() &&
  github.event_name != 'pull_request' &&
  needs.preflight.outputs.should_build == 'true' &&
  contains(fromJSON('["success", "failure"]'), needs.buildroot.result)
```

So producing `fw-*` on a pull request cannot reach a release. It only makes the
image downloadable from the run that built it. `Send binary` keeps its guard
too, so no PR touches the Telegram secret or the dev channel.

## What it costs

`select` narrows the matrix via `ci-matrix.py`, and the cost depends on what the
PR touches — measured, not estimated:

| PR shape | devices built | of 96 |
| --- | --- | --- |
| device profile (#119) | **1** | 1% |
| CI infrastructure (this PR) | **15** | 16% |

So the case this change exists for — a contributor adding a device — produces a
single ~7 MB artifact. A workflow change like this one produces 15, roughly
100 MB at 7-day retention. Both are modest, but the second is not "one image per
PR" and I would rather state it than round it down.

## Retention, per event

```yaml
retention-days: ${{ github.event_name == 'pull_request' && 7 || 1 }}
```

A nightly's artifact is consumed by `publish` inside the same run and nobody
downloads it afterwards, so a nightly gains nothing from a longer window — and
since it builds the whole matrix daily, a week of retention would keep seven
full matrices in storage instead of one, for no reader. A PR builds a handful
and has a person waiting on them, so one day is too short there.

The first revision of this PR set `7` unconditionally, which contradicted that
reasoning; corrected in b14cc52 after review.

## Verification

The change is two `if:` simplifications and one retention bump; the guards that
matter are untouched. `.github/scripts/lint-workflow-shell.py` passes — 31 run
blocks, all clean.

The honest test is this PR's own run: **`Upload build artifacts` should now
succeed and attach `fw-*`, while `Publish releases` stays skipped.** If both hold
here, the change does what it claims. Worth confirming that before merge rather
than taking the reasoning for it.

### DIFF
--- a/.github/workflows/master.yml
+++ b/.github/workflows/master.yml
@@ -274,10 +274,14 @@ jobs:
       # schemes the build used (compound devices rename to <device>-nor.tgz in
       # the workspace root, simple ones leave openipc.<soc>-nor-<variant>.tgz
       # under openipc/output/images/).
+      # Runs on pull_request too. Staging and the upload below are run-scoped
+      # and touch no secret and no shared state, so unlike `Send binary` and the
+      # `publish` job -- which stay gated, and must -- there is nothing here for
+      # a PR to leak or clobber. Without this a contributor's PR compiles a full
+      # image and then discards it, so they cannot test their own device profile
+      # without asking a maintainer to dispatch build-one for them.
       - name: Stage artifacts
-        if: >-
-          github.event_name != 'pull_request' &&
-          (env.NORFW || env.NANDFW)
+        if: env.NORFW || env.NANDFW
         run: |
           mkdir -p dist
           for f in "${NORFW:-}" "${NANDFW:-}" "${SIZES:-}"; do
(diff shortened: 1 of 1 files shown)


widgetii:
Fixed in b14cc52 — the finding was correct, and it caught a real inconsistency
rather than a style point.

The rationale I wrote for the change argued that a nightly's artifact is consumed
by `publish` inside the same run and never needs to outlive it. Having argued
that, I then set `retention-days: 7` unconditionally, which applies the longer
window to precisely the case that has no use for it — and since a nightly builds
the entire device matrix daily, that is seven full matrices held in storage
instead of one, for no reader at all. The code contradicted its own comment.

Now per-event:

```yaml
retention-days: ${{ github.event_name == 'pull_request' && 7 || 1 }}
```

`pull_request` → 7, every other event → 1, unchanged from today's behaviour.
Checked the expression against the `A && B || C` idiom for each trigger the
workflow declares:

| event | retention-days |
| --- | --- |
| `pull_request` | 7 |
| `schedule` | 1 |
| `workflow_dispatch` | 1 |
| `push` | 1 |

`.github/scripts/lint-workflow-shell.py` still passes.

Also worth recording that the first revision already demonstrated the change
works: on the run for `7530413`, the first device to finish reported
`Stage artifacts` **success**, `Upload build artifacts` **success**, and
`Send binary` **skipped**, with `Publish releases` not running. That is the
behaviour this PR claims — the image is kept, and nothing secret-backed or
release-writing is reached from a pull request.

PR #138, opened 2026-08-30
publish with an explicit release credential, and prove it before building

openipc-ai:
Companion to OpenIPC/firmware#2339. This repo was **next in line** for the same failure and hasn't hit it yet only because the release train correctly refused to advance past firmware's failure.

## The failure being pre-empted

Moving the nightly off `schedule:` onto a dispatch changes the identity the run executes as, and the default `GITHUB_TOKEN`'s release-write goes with it. For an app/OAuth-backed identity the releases API answers:

```
HTTP 403: Resource not accessible by integration
```

even though the job declares `contents: write` and the run log confirms the token was *granted* it — which is why nothing in the workflow file looks wrong. OpenIPC/firmware's run 33273333359 built 102 boards over two hours before finding out; `latest` never moved and cameras stopped updating.

This repo publishes exactly the same way — `gh release create` plus a PATCH of `refs/tags/latest` — so it would have failed identically, after building all 107 devices.

## Changes

**1. `publish` uses `secrets.RELEASE_TOKEN`** (`Contents: read and write`). Also makes publishing independent of who or what starts the run.

**2. `preflight` proves the credential first** — draft release created and deleted; invisible, no tag, two API calls. Checking the secret is merely non-empty would not have caught the firmware incident: that token existed and was simply refused, so the probe attempts the write.

Skipped on `pull_request`, which neither publishes nor has the secret.

## Before merging

`RELEASE_TOKEN` must exist as a repo secret here, or the next non-PR run fails fast in preflight — intended, but red rather than silently unpublished.

## Verification

- Workflow YAML parses; `.github/scripts/ci-matrix.py --self-test` passes (111 devices, 15 smoke, 39 cases).
- Both probe failure paths exercised locally: unset token and refused API call each produce a clear `::error::` and exit 1. Neither passes silently.

### DIFF
--- a/.github/workflows/master.yml
+++ b/.github/workflows/master.yml
@@ -129,6 +129,43 @@ jobs:
           echo "build_id=$BUILD_ID"  >> "$GITHUB_OUTPUT"
           echo "built_at=$BUILT_AT"  >> "$GITHUB_OUTPUT"
 
+      # Prove the release credential BEFORE the matrix runs.
+      #
+      # OpenIPC/firmware run 33273333359 built 102 boards over two hours and
+      # then died in `Publish releases` with "403 Resource not accessible by
+      # integration". Nothing earlier in that run had any opinion about whether
+      # it could publish, so the price of the misconfiguration was the whole
+      # matrix. This repo builds 107 devices and publishes the same way.
+      #
+      # A draft release is not visible to anyone and creates no tag, so this
+      # costs two API calls and turns a long failure into a ten-second one.
+      # Checking the secret is merely non-empty would not have caught that
+      # incident: the old token existed and was simply refused.
+      #
+      # Skipped on pull_request, which neither publishes nor has the secret.
+      - name: Prove the release credential can write
+        if: github.event_name != 'pull_request'
+        env:
+          GH_TOKEN: ${{ secrets.RELEASE_TOKEN }}
+          GH_REPO: ${{ github.repository }}
+        run: |
+          set -euo pipefail
+          if [ -z "${GH_TOKEN:-}" ]; then
+            echo "::error::RELEASE_TOKEN is not set. The matrix would build"\
+                 "every device and then fail to publish. See the publish job."
+            exit 1
+          fi
+          probe="ci-release-credential-probe-${{ github.run_id }}"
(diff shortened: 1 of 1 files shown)


openipc-ai:
Both findings addressed. The transient-failure one was right, and sharper than it looks.

### 1. Probe misclassifies transient failures — fixed in `0f0229d`

Correct, and I'd already seen evidence for it without connecting it: listing releases in these repos returned **HTTP 504** twice while I was developing this change. GitHub also answers **403 for secondary rate limiting**, which is precisely why the publish step below retries six times. So a one-shot POST reporting "RELEASE_TOKEN cannot create releases" would send whoever read that log off to reissue a credential that was never the problem — a misdiagnosis baked into the error text.

The POST now retries 3× with 5s/10s backoff, and the message reads *"could not create a release after 3 attempts. If this is not a transient API error, the token needs Contents: read and write."*

Three rather than the publish step's six on purpose: the value of this probe is failing **fast**. A genuine misconfiguration still surfaces in well under a minute instead of after 107 devices have built.

### 2. Probe drafts leak on interruption — fixed in `c5930f9`

Already addressed before this review landed (Qodo marks it resolved). Tag now carries `run_attempt` as well as `run_id` — `run_id` is stable across re-runs, so a leftover would otherwise collide with its own re-run and report a valid credential as broken. Deletion moved into a `trap`, plus a best-effort sweep for drafts stranded by earlier runs, since `cleanup.yml` doesn't know this tag pattern.

### A third one, found by testing the fix for the first

Exercising the retry path surfaced a defect in that sweep: `gh` writes its error body to **stdout**, so a failed list fed the cleanup loop lines of JSON instead of release ids, and the log filled with:

```
removing stranded probe draft {"status": "401"}
```

The deletes were harmless — they simply failed — but a step whose entire purpose is to be believed must not narrate confident nonsense. Only numeric ids are acted on now.

### Verification

Against the live repo, not just reasoned about:

- happy path → creates draft, `release credential OK`, trap removes it, **0 drafts left**, no tag created, published releases untouched
- refused credential → retries 3×, fails with the honest message, exit 1
- failing list → **0** bogus removal lines

PR #151, opened 2026-09-09
Add t31_lite_zte-k545 device profile (ZTE K545, SC4336, ATBM6012B Wi-Fi)

tagbug:
Adds a device profile for the ZTE K545 PTZ camera — Ingenic T31X, SC4336 sensor, 16 MiB NOR, and ATBM6012B USB Wi-Fi.

The K545 shares part of its hardware setup with the existing ZTE K540 profile, but requires its own GPIO/PTZ configuration:

* Reuses the proven `atbm603x-t31-zte-k540` Wi-Fi runtime profile.
* Uses GPIO 49/63/62/61/59/52/53/54 for the pan/tilt motors.
* IR-cut uses GPIO 57/58, with the IR backlight on GPIO 11.
* Speaker enable is GPIO 64, active low.
* MicroSD card detect uses PB18 / GPIO 50, carrying the same MMC board fix as the existing K540 profile.
* The board has 16 MiB NOR, while the profile intentionally keeps the same 8 MiB OpenIPC image/partition footprint used by K540.

Tested on a physical ZTE K545: the firmware builds, flashes and boots successfully, the device operates normally, and the Majestic WebUI works correctly.

All repository checks pass (`ci-matrix`, `check-firmware-drift --self-test`, `lint-cli-paths`, and the relevant repository self-tests).


openipc-ai (commented):
Thanks — and sorry this sat unbuilt for two days: a PR from a first-time contributor needs a maintainer to release the workflow run. Reviewed against master `e0a643f` with firmware at `9e7cf5b`; nothing here blocks, so I am releasing the build now.

It is a faithful clone of the in-tree `t31_lite_zte-k540` with the deltas the K545 actually needs, which is exactly the recipe. Things I checked rather than assumed:

- `fw_setenv wlandev atbm603x-t31-zte-k540` reuses the K540 profile, which does exist upstream in firmware's `general/overlay/etc/wireless/usb`. Reusing it is fine as long as the radio and its power-up are wired the same; if the K545 ever needs its own sequence it wants its own profile name.
- The kernel patch is byte-identical to K540's, and it will be applied: `general/openipc.fragment` sets `BR2_GLOBAL_PATCH_DIR="$(BR2_EXTERNAL)/package/all-patches"`, and `builder.sh` copies your `devices/<dev>/general/package/all-patches/` over that directory. `GPIO_PB(18)` is pin 50, so the filename and the hunk agree.
- Moving to `BR2_PACKAGE_GPIO_MOTORS` with `ptz_control`/`ptz_gpio` rather than K540's `BR2_PACKAGE_MOTORS` is the current pattern — good.
- `ci-matrix.py --stdin` on your branch narrows to exactly one device, so this costs a single build.

Two small things, neither blocking:

1. **`BR2_OPENIPC_FLASH_SIZE="8"` but the README row says `NOR_16M`.** K540 carries the same mismatched pair, so you have probably inherited it rather than introduced it — but if the K545 really has 16M, this leaves half the flash unused. Which is right?
2. `# BR2_PACKAGE_QUIRC_OPENIPC=y` is a plain comment, not a kconfig "off" line. The effect is the same (unset), but two lines down you use the canonical `# BR2_PACKAGE_VTUND_OPENIPC is not set`, so the file disagrees with itself. Worth making it `# BR2_PACKAGE_QUIRC_OPENIPC is not set` — note that also drops the QR-code Wi-Fi provisioning K540 ships, which is intentional if you meant it.

Let's see what the build says.

PR #152, opened 2026-09-09
Add ssc337de_lite_zte-k543 device profile (ZTE K543, SC401AI, MT7601U Wi-Fi)

tagbug:
Adds a device profile for the ZTE K543 PTZ camera — SigmaStar SSC337DE (Infinity6B0), SC401AI sensor, 16 MiB NOR, and MT7601U USB Wi-Fi.

The board needs a few device-specific settings beyond the standard SSC337DE lite configuration:

* MT7601U initialization requires a GPIO 14 toggle before loading `mt7601sta`; the device-specific wireless overlay implements this and is tracked in `firmware-drift.json`.
* Pan/tilt motors use GPIO 0-7 via `gpio-motors`.
* IR-cut uses GPIO 79/78, with the IR backlight on GPIO 52.
* Speaker enable is GPIO 15, active low.
* The profile selects SC401AI explicitly and excludes unused sensor/NFS kernel modules.

Tested on a physical ZTE K543: the firmware builds, flashes and boots successfully; SC401AI video, MT7601U Wi-Fi, IR-cut/night mode, speaker output, and both PTZ axes are working.

All repository checks pass (`ci-matrix`, `check-firmware-drift --self-test`, `lint-cli-paths`, `lint-workflow-shell`, and shell syntax validation).


openipc-ai (commented):
Thanks — and sorry this sat unbuilt for two days: a PR from a first-time contributor needs a maintainer to release the workflow run. Reviewed against master `e0a643f` with firmware at `9e7cf5b`; nothing here blocks, so I am releasing the build now.

This is unusually well put together for a first device PR. Three things I verified rather than took on trust:

- **The `firmware-drift.json` entry is exactly right.** `f9df96bc190e8f8a11794f9410568eac10f50320` is the current blob of firmware's `general/overlay/etc/wireless/usb` at HEAD — I resolved it independently and it matches. Most contributors do not know that file exists; registering the overlay there is the difference between a tracked divergence and a silent one.
- **The exclude list keeps the right sensor.** `OpenIPC/sensors` builds 18 sensors for `sigmastar/infinity6b0`, your list prunes 17 of them and leaves `sc401ai` — the board's own — in place.
- `mt7601sta` is the right module name: `mt7601u-openipc` builds the vendor STA driver (it installs `/etc/mediatek/MT7601USTA.dat`), and three existing upstream profiles modprobe that same name. The kernel config you point at, `infinity6b0-ssc009b.config`, is the one the other ssc337de device in the tree uses.

One nit: `/lib/modules/4.9.84/sigmastar/sensor_jxq03_mipi.ko` is not built for infinity6b0 — that sensor exists for other families but not this one — so the entry will never match anything, and `rootfs_script.sh` will count it in its `excludes: N of M entries matched no file` line. Harmless, just drop the line.

`ci-matrix.py --stdin` on your branch narrows to exactly one device, so this is a single build. Let's see what it says.

PR #154, opened 2026-09-11
t31_lite_zte-k545: ship the excludes list the device was missing

openipc-ai:
#151 added the K545 without `general/scripts/excludes/t31_lite.list`, one of the files README lists as required for a registered device. I merged it anyway and offered to follow up, which was the wrong call — this is that fix.

Nothing failed for the want of it: the image builds and the author has the camera running. But it builds close to the edge. From #151's own CI run:

```
- uImage:          [1786KB/2048KB]
- rootfs.squashfs: [4560KB/5120KB]
  headroom: kernel=261KB rootfs=560KB
```

Every sensor the T31 osdrv ships is in that rootfs, and the board has exactly one (SC4336).

The list is `t31_lite_chinatelecom-y4h-50`'s, which is the conservative of the two in the tree: it prunes other sensors' `.bin`, `.yaml` and `.ko` and stops there. The `tp-link-tapo-tc70-v3` variant additionally globs `kernel/drivers/net/wireless/*`, and this board's Wi-Fi is the out-of-tree atbm60xx module, so that one is not worth the risk here for the few KB it would add.

Checked rather than assumed: the osdrv ships `sc4336-t31.bin` and `sc4336.yaml`, and neither appears in the list, so the K545's own sensor survives the prune. `ci-matrix.py --stdin` narrows this to the one device, so the build in this PR is the measurement — the size report will say how much headroom it bought.

The sibling `t31_lite_zte-k540` has the same gap and the same sensor. Left alone here for its author to decide.

On the other thing I raised at merge time, the 8M `BR2_OPENIPC_FLASH_SIZE` against a `NOR_16M` README row: that one is not a defect, and I should not have asked. #151's description already says it — "The board has 16 MiB NOR, while the profile intentionally keeps the same 8 MiB OpenIPC image/partition footprint used by K540."


openipc-ai:
Closing this: the build went green and then told me the change does nothing.

```
excludes: 65 of 65 entries in t31_lite.list matched no file
- rootfs.squashfs: [4564KB/5120KB]   (4560KB before this PR)
```

The premise was wrong. `ingenic-osdrv-t31.mk` branches on `OPENIPC_SNS_MODEL`:

```make
ifeq ($(OPENIPC_SNS_MODEL),)
    ... installs every *.yaml and every params/*-t31.bin
else
    ... installs $(OPENIPC_SNS_MODEL).yaml and $(OPENIPC_SNS_MODEL)-$(OPENIPC_SOC_MODEL).bin
endif
```

The K545 sets `BR2_OPENIPC_SNS_MODEL="sc4336"`, so its image never contained the other sensors' files in the first place — there was nothing for the list to prune, and the 4KB difference is squashfs noise.

That also explains the three t31_lite devices that do ship this list: `tp-link-tapo-tc70-v3`, `chinatelecom-y4h-50` and `aoni-ep01j05` all leave `BR2_OPENIPC_SNS_MODEL` empty, so for them the osdrv installs all 18 sensors and the list earns its place. Pinning the sensor model and shipping an excludes list are two routes to the same end, and the K545 already takes the first one.

So the rootfs sits at 4564KB/5120KB because of majestic, the webui and the kernel modules, not because of stray sensor assets, and this file would only have added a permanent "65 of 65 entries matched no file" line to every build of this device.

Worth noting for the next reader: the automated review on #151 flagged the missing list as a rule violation, and README does list it among a device's required files. For a device that pins `BR2_OPENIPC_SNS_MODEL`, that is not true — which is a docs bug rather than a device bug.

PR #156, opened 2026-09-12
Switcam B200

Puzzak01:


devices/ssc337de_ultimate_switcam-b200/general/scripts/excludes/ssc337de_ultimate.list
```diff
@@ -0,0 +1,5 @@
+/lib/modules/4.9.84/kernel/fs/lockd/lockd.ko
```
openipc-ai, line 1: **This list will never be read — it needs the flavor in its name.**

The filename is computed by the firmware, not taken from the device directory. `general/scripts/rootfs_script.sh:25`:

```sh
LIST="${BR2_EXTERNAL_GENERAL_PATH}/scripts/excludes/${OPENIPC_SOC_MODEL}_${OPENIPC_VARIANT}.list"
```

For this board that resolves to `ssc337de_ultimate.list`, so `ssc337de.list` is never opened. And the whole block is wrapped in `if [ -f "${LIST}" ]` — the script reports, never fails — so a misnamed list prunes nothing and prints no diagnostic at all. Rename the file to `ssc337de_ultimate.list` (keep it named after soc+flavor, not the device).

While you are in here: this list carries only the five nfs/sunrpc entries, where the sibling `devices/ssc337de_lite_zte-k543/general/scripts/excludes/ssc337de_lite.list` also prunes 18 `sensor_*_mipi.ko`. Worth mirroring that block minus `sensor_sc223a_mipi.ko`, which is the sensor this board actually uses.

One thing not to rely on as a substitute: pinning `BR2_OPENIPC_SNS_MODEL="sc223a"` in the defconfig will not cover those modules. `general/package/sigmastar-osdrv-infinity6b0/sigmastar-osdrv-infinity6b0.mk` branches on `OPENIPC_SNS_MODEL` only for `/etc/sensors/*.bin`; the `sensor_*_mipi.ko` come from the kernel build and are unaffected. (This is the opposite of T31, where pinning the sensor does make the list redundant.)

Once the list is being read you can confirm it in the build log — `rootfs_script.sh` prints `excludes: N of M entries in <name> matched no file` for a list it opened, which is the cheap way to tell a working list from a decorative one.

devices/ssc337de_ultimate_switcam-b200/general/overlay/usr/share/openipc/customizer.sh
```diff
@@ -0,0 +1,26 @@
+#!/bin/sh
+#
+# Perform basic settings on a known IP camera
+#
+#
+# Set custom upgrade url
+#
+fw_setenv upgrade 'https://github.com/OpenIPC/builder/releases/download/latest/ssc337de_ultimate_switcam-b200-nand.tgz'
```
openipc-ai, line 8: **`-nand.tgz` needs confirming — it is only produced under one condition.**

A nand artifact exists only when the defconfig sets `BR2_TARGET_ROOTFS_UBI=y`; see the `repack` target in the firmware `Makefile`:

```make
ifeq ($(BR2_TARGET_ROOTFS_UBI),y)
ifneq ($(filter $(BR2_OPENIPC_SOC_VENDOR),"rockchip" "sigmastar"),)
	@$(call PREPARE_REPACK,,,rootfs.ubi,16384,nand)
```

CI then renames it to `<device>-nand.tgz`, so the name here is right *if* the build emits one — which cannot be checked, because the defconfig is missing.

Worth a deliberate answer rather than a default: every NAND board currently in the tree points its upgrade URL at `-nor.tgz`, not `-nand.tgz` — `hi3516ev300_ultimate_rostelecom-ipc8232swc-we`, `hi3516ev300_ultimate_rvi-1ncmw2028`, and `ssc325de_lite_imou-c22ep-s2` (the closest match: sigmastar, `BR2_TARGET_ROOTFS_UBI=y`). They all set `BR2_TARGET_ROOTFS_SQUASHFS=y` alongside UBI and so build both artifacts.

So either set `BR2_TARGET_ROOTFS_UBI=y` and keep `-nand.tgz` deliberately, or match the siblings. Please say which, so whoever hits a broken self-update later can find the reasoning.

(The Qodo bot above states `-nor.tgz` is mandatory — that is not true in general, the suffix follows the artifact the build actually produces.)

README.md
```diff
@@ -60,6 +60,7 @@ Smartwares CIP-37210     HI3518EV200  OV9732    RTL8188FU_USB    NOR_16M   done
 Smartwares CIP-37210AT   T21N         JXF37     RTL8188FU_USB    NOR_16M   done
 Switcam HS303 v1         HI3518EV200  JXF22     RTL8188FU_USB    NOR_16M   done
 Switcam HS303 v2         HI3518EV200  OV9732    RTL8188EU_USB    NOR_16M   done
+Switcam B200             SSC337DE     SC223A                     NAND      done
```
openipc-ai, line 63: Small one: the table is sorted alphabetically, so `Switcam B200` belongs *above* the two `Switcam HS303` rows, not below them. Column alignment itself is correct.

Also worth confirming rather than leaving blank: the WiFi column is empty and the customizer sets no `fw_setenv wlandev`. Those two are at least consistent with each other, but an ethernet-only Switcam is unusual — if the board does have a WiFi module, both this row and the defconfig's `BR2_PACKAGE_<chip>_OPENIPC=y` line need it.

openipc-ai (changes requested):
Thanks for the profile — the runtime side (GPIO map, night-mode pins, audio) looks like real hardware work. But the device is not registered, so nothing here is reachable by a build yet.

## Blocking: there is no defconfig

`devices/ssc337de_ultimate_switcam-b200/` ships only `general/`. The file that registers a device is missing:

```
devices/ssc337de_ultimate_switcam-b200/br-ext-chip-sigmastar/configs/ssc337de_ultimate_switcam-b200_defconfig
```

Its name must match the directory name exactly — that is what `make BOARD=` looks up, and `./builder.sh ssc337de_ultimate_switcam-b200` fails without it.

The consequence is worse than "the board is not built", and it is visible on this PR right now:

```console
$ python3 .github/scripts/ci-matrix.py --self-test     # master: 114 devices
$ python3 .github/scripts/ci-matrix.py --self-test     # this branch: 114 devices

$ gh pr view 156 --json files -q '.files[].path' | python3 .github/scripts/ci-matrix.py --stdin
ci-matrix: 114/114 devices (needs_build=True) --- devices/ssc337de_ultimate_switcam-b200/general/overlay/usr/share/openipc/customizer.sh is in no directory CI builds
```

Two things fall out of that. The self-test passes, because a directory with no defconfig is not in the device set at all — there is nothing for it to disagree about, so CI cannot warn you. And because `ci-matrix.py` maps a changed file to "the devices whose defconfig shares its directory", an unrecognised path *widens* back to the full matrix instead of narrowing. Run 34712449145 on this branch is therefore building all 114 devices, none of which is the B200.

The closest template is `devices/ssc337de_ultimate_foscam-x5/` — same SoC, same `ultimate` flavor. Copy its defconfig and change the SoC/flash/WiFi/sensor lines for this board. Note `BR2_OPENIPC_FLASH_SIZE` is the NOR/NAND size in MB and is still required.

## Also

Three inline comments below: the excludes filename, the `-nand.tgz` artifact name, and the README row position.

Happy to help with the defconfig if you can post the board's flash size, WiFi module (if any) and NAND geometry — those are the parts only someone with the hardware can supply.

openipc-ai:
Thanks for the quick turnaround — two of the three are properly fixed:

- `ssc337de_ultimate.list` is now the name `rootfs_script.sh` actually looks up, so the list will be read.
- `BR2_TARGET_ROOTFS_UBI=y` (plus `UBI_SUBSIZE`, `ubinize_sigmastar.cfg`, `LEBSIZE=0x1f000`) means the build really does emit a nand artifact, so `-nand.tgz` in the customizer is correct rather than aspirational. Matching `ssc325de_lite_imou-c22ep-s2` there was the right call.

The blocker is still open, though — and for a reason you cannot see anywhere in the GitHub UI.

## The defconfig filename ends in an invisible character

`ssc337de_ultimate_switcam-b200_defconfig` has a stray **U+200E LEFT-TO-RIGHT MARK** appended to it:

```console
$ ls devices/ssc337de_ultimate_switcam-b200/br-ext-chip-sigmastar/configs/ | cat -A
ssc337de_ultimate_switcam-b200_defconfigM-bM-^@M-^N$
                                         ^^^^^^^^^^  e2 80 8e = U+200E
```

It renders as nothing — in the file tree, in the diff, in the commit, in your editor. But the filename no longer equals the directory name, which is the exact string `make BOARD=` and `ci-matrix.py` look for, so nothing has changed since the last push:

```console
$ python3 .github/scripts/ci-matrix.py --self-test
ci-matrix: self-test ok (114 devices, ...)          # identical to master -- device not registered

$ gh pr view 156 --json files -q '.files[].path' | python3 .github/scripts/ci-matrix.py --stdin
ci-matrix: 114/114 devices (needs_build=True) --- devices/ssc337de_ultimate_switcam-b200/br-ext-chip-sigmastar/configs/ssc337de_ultimate_switcam-b200_defconfig<U+200E> is in no directory CI builds
```

My guess is it rode along in a pasted filename — invisible marks like this are easy to pick up copying a name out of a rendered page or a chat client.

## The fix

Deleting that one character is the whole fix. Verified locally on your branch:

```console
$ python3 .github/scripts/ci-matrix.py --self-test
ci-matrix: self-test ok (115 devices, ...)

$ gh pr view 156 --json files -q '.files[].path' | python3 .github/scripts/ci-matrix.py --stdin
ci-matrix: 1/115 devices (needs_build=True) --- narrowed to the affected devices
  ssc337de_ultimate_switcam-b200
```

114 -> 115 devices, and the PR stops building the entire matrix and builds just your board.

Locally:

```sh
cd devices/ssc337de_ultimate_switcam-b200/br-ext-chip-sigmastar/configs
git mv "$(ls)" ssc337de_ultimate_switcam-b200_defconfig
git commit -m "strip stray U+200E from defconfig filename"
```

Or in the web UI: open the file, hit the pencil, clear the filename box completely and **retype** the name by hand rather than pasting it. Then confirm with `ci-matrix.py --self-test` -- it should say 115 devices, and that is the check CI runs too.

Happy to push the rename to `patch-3` myself if you would rather not fight an invisible byte -- `maintainer_can_modify` is on. Just say the word.

Still open from the earlier review, neither blocking:

- The excludes list is still only the five nfs/sunrpc entries. Mirroring the sensor block from `ssc337de_lite.list` (minus `sensor_sc223a_mipi.ko`, which this board uses) is what the list is for, though at 16M NAND you have the room either way.
- The README row sorts below the two `Switcam HS303` rows; alphabetically `B200` goes above them.

PR #159, opened 2026-09-14
hi3518ev100: add HiWatch DS-I203 device profile

ArthurKoba:
## Summary

Adds an OpenIPC device profile for the **HiWatch DS-I203** camera.

Tested hardware:

- HI3518EV100
- Sony IMX122
- 256 MiB DDR3
- GD25Q128 16 MiB SPI NOR
- RTL8201F Ethernet PHY at address 3

The 256 MiB physical memory is configured as **128 MiB Linux + 128 MiB MMZ**.

This hardware requires the DDR3/256M U-Boot variant added in [OpenIPC/u-boot-hi3516cv100#6](https://github.com/OpenIPC/u-boot-hi3516cv100/pull/6).

## Device configuration

The profile provides the board-specific configuration required by the DS-I203:

- `osmem=128M`
- 16 MiB NOR partition layout
- PHY address 3 / MDIO interface 0
- IMX122 sensor selection
- `/etc/sensors/imx122_spi_dc_1080p.ini`
- IR-cut control on GPIO 6/5
- IR backlight on GPIO 42

The first-boot customizer writes the required U-Boot environment and performs a delayed reboot so the memory layout and Ethernet configuration take effect on the next boot.

The IR illumination hardware also supports PWM brightness control and PWM0 was verified working on the camera. The current `hi3516cv100` Majestic HAL does not provide PWM backlight control, so the profile uses GPIO ON/OFF control for now.

Native PWM support is tracked in [OpenIPC/majestic#321](https://github.com/OpenIPC/majestic/issues/321).

## Hardware validation

Tested end-to-end on a physical HiWatch DS-I203, starting from the original Hikvision firmware and completing the full migration to OpenIPC.

After the first-boot customizer and automatic reboot:

- U-Boot detects 256 MiB RAM
- Linux boots with `mem=128M`
- MMZ is 128 MiB
- the 16 MiB MTD layout is correct
- Ethernet works with PHY address 3 / MDIO interface 0
- the factory MAC address is preserved
- IMX122 is selected with the expected sensor configuration
- Majestic starts successfully
- IR-cut control works
- IR backlight control works

Final firmware build size:

- kernel: 1819 KiB / 2048 KiB
- rootfs: 4056 KiB / 8192 KiB

The full device build and repository checks pass.


devices/hi3518ev100_lite_hiwatch-ds-i203/br-ext-chip-hisilicon/configs/hi3518ev100_lite_hiwatch-ds-i203_defconfig
```diff
@@ -0,0 +1,58 @@
...
+BR2_PACKAGE_UBOOT_TOOLS=y
+BR2_PACKAGE_ZLIB=y
+# BR2_PACKAGE_IFUPDOWN_SCRIPTS is not set
+BR2_TARGET_ROOTFS_CPIO=y
+BR2_TARGET_ROOTFS_SQUASHFS=y
+BR2_TARGET_ROOTFS_SQUASHFS4_XZ=y
+
+# OpenIPC
+BR2_OPENIPC_SOC_VENDOR="hisilicon"
+BR2_OPENIPC_SOC_MODEL="hi3516cv100"
```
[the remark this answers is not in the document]
ArthurKoba: Fixed. `BR2_OPENIPC_SOC_MODEL` is now `hi3518ev100`, while `BR2_OPENIPC_SOC_FAMILY` remains `hi3516cv100` as required by the shared SDK family. I also aligned the kernel config path with the upstream `hi3518ev100_lite_defconfig`.

devices/hi3518ev100_lite_hiwatch-ds-i203/general/overlay/usr/share/openipc/customizer.sh
```diff
@@ -0,0 +1,28 @@
+#!/bin/sh
+
+# Requires u-boot-hi3518ev100-ddr3-256m-universal.bin.
+# The generic HI3518EV100 U-Boot does not match this board's DDR configuration.
+fw_setenv upgrade 'https://github.com/OpenIPC/builder/releases/download/latest/hi3518ev100_lite_hiwatch-ds-i203-nor.tgz'
+
+# Board-specific U-Boot environment.
+fw_setenv osmem 128M
```
openipc-ai, line 8: **This needs `fw_setenv totalmem 256M` next to it, or the camera loses video on the very next boot.**

`hisilicon-osdrv-hi3516cv100/files/script/load_hisilicon` opens with:

```sh
mem_total=$(fw_printenv -n totalmem | tr -d 'M')
mem_total=${mem_total:=64}
```

and ends the memory block with a hard guard:

```sh
if [ $os_mem_size -ge $mem_total ]; then
	echo "[err] os_mem[$os_mem_size], over total_mem[$mem_total]"
	exit
fi
```

`calc_mmz_info` then derives `mmz_size = totalmem - osmem`, so `totalmem` is the only source of the MMZ size — its absence silently becomes 64.

The trap is that `totalmem` is not reliably in the saved environment. OpenIPC's u-boot sets it from `detect_memory()` in `board/hi3518/board.c`, but that is a plain `setenv` into the RAM copy, never a `saveenv`. `fw_printenv` reads the env partition in flash, so it sees `totalmem` only if some install step happened to persist it. `run setnor16m` does, by accident, because it ends in `saveenv` — whereas the wiki's `en/install-hisi.md` procedure has the user type `setenv totalmem 64M` before `saveenv`, which persists the wrong value on any board with more than 64 MiB.

Boot by boot on a fresh install:

- **Boot 1** — env still has u-boot's default `osmem=32M`, so `os_mem_size=32`, `32 >= 64` is false, modules load with a 32/32 split. This customizer runs and reboots.
- **Boot 2** — cmdline is now `mem=128M`, `os_mem_size` resolves to 128, `128 >= 64` trips, `load_hisilicon` exits before `insert_osal`. No `mmz.ko`, no `hi3518_*.ko`, majestic cannot open the SDK. Permanent.

Your camera presumably has `totalmem=256M` in flash from the `saveenv` during bring-up, which is exactly why it worked for you and will not work for the next person. Pinning it makes the profile deterministic:

```sh
fw_setenv osmem 128M
fw_setenv totalmem 256M
```

`devices/gk7205v300_lite_vixand-ivg-g6s` and `devices/hi3516ev300_ultimate_rostelecom-ipc8232swc-we` already set both together for the same reason.

devices/hi3518ev100_lite_hiwatch-ds-i203/general/overlay/usr/share/openipc/customizer.sh
```diff
@@ -0,0 +1,28 @@
...
+fw_setenv mtdparts 'hi_sfc:256k(boot),64k(env),3072k(kernel),10240k(rootfs),-(rootfs_data)'
```
openipc-ai, line 9: **Please drop this line.** Rewriting the flash layout from a first-boot script is either a no-op or a brick, depending on how the camera was installed — and there is no way for the script to tell which.

The string itself is correct: it is byte-identical to `mtdpartsnor16m` in u-boot's `include/configs/hi-common.h`. But that is an opt-in macro, not the default. The default in the same `CONFIG_EXTRA_ENV_SETTINGS` block is the 8 MiB layout:

```
mtdparts       = hi_sfc:256k(boot),64k(env),2048k(kernel),5120k(rootfs),-(rootfs_data)   <- default
mtdpartsnor16m = hi_sfc:256k(boot),64k(env),3072k(kernel),10240k(rootfs),-(rootfs_data)  <- only via `run setnor16m`
```

So there are two install paths:

- User ran `run setnor16m` (whose `saveenv` persists the 16M string) and flashed with `uknor16m`/`urnor16m` — rootfs at 0x350000. The env already holds this exact string and the line does nothing.
- User flashed with the 8M macros — `urnor8m` writes the rootfs at 0x250000. That succeeds on a 16 MiB chip, because a 4056 KiB rootfs fits the 5120k window fine. The camera boots, reaches `S30customizer`, and this line moves `root=/dev/mtdblock3` from 0x250000 to 0x350000. Three seconds later `reboot -f` fires into blank flash: kernel panic, recoverable only over UART.

Flash layout belongs to the install procedure, not to a first-boot script — no other device under `devices/` sets `mtdparts`. If the DS-I203 needs the 16M layout, that belongs in the install instructions as `run setnor16m` before flashing.

devices/hi3518ev100_lite_hiwatch-ds-i203/general/overlay/usr/share/openipc/customizer.sh
```diff
@@ -0,0 +1,28 @@
...
+cli -s .nightMode.irCutPin1 6
+cli -s .nightMode.irCutPin2 5
+
+cli -s .nightMode.backlightEnabled true
+cli -s .nightMode.backlightPin 42
+cli -s .nightMode.lightMonitor true
+
+
+# Reboot once so the updated boot-time environment takes effect.
+(sleep 3 ; reboot -f) &
```
openipc-ai, line 26: Follow-on from the `mtdparts` line above, in the branch where it does change something: `rootfs_data` moves too, from 256+64+2048+5120 = 7488 KiB (0x750000) to 256+64+3072+10240 = 13632 KiB (0xD50000).

`S30customizer` runs this script and then does `touch /etc/custom.ok`, both landing in the overlay at the old offset. After `reboot -f` the overlay is mounted at the new one, unformatted — so `.isp.sensorConfig`, the IR-cut pins, the backlight pin and `lightMonitor` are all gone, `/etc/custom.ok` is gone, and the customizer runs again and reboots again. Dropping the `mtdparts` line fixes this one too.

Separately, and not something this PR introduced: `(sleep 3 ; reboot -f) &` returns immediately, so `rcS` carries on into `S70vendor` (`load_hisilicon -i`, a serial insmod of the vendor modules) while the timer runs. On a slow first boot the reboot can land in the middle of that. Several existing profiles have the same race, so I would not block on it here.

devices/hi3518ev100_lite_hiwatch-ds-i203/general/overlay/usr/share/openipc/customizer.sh
```diff
@@ -0,0 +1,28 @@
+#!/bin/sh
+
+# Requires u-boot-hi3518ev100-ddr3-256m-universal.bin.
```
openipc-ai, line 3: Small thing: this note sits above the `upgrade` URL, which has nothing to do with the bootloader. The line it actually explains is `fw_setenv osmem 128M` below — 128M is only valid on the DDR3/256M u-boot variant. Worth moving it down two lines (and above the `totalmem` line it should gain), so someone editing the self-update URL later does not read it as a constraint on that.

devices/hi3518ev100_lite_hiwatch-ds-i203/general/overlay/usr/share/openipc/customizer.sh
```diff
@@ -0,0 +1,28 @@
...
+# The generic HI3518EV100 U-Boot does not match this board's DDR configuration.
+fw_setenv upgrade 'https://github.com/OpenIPC/builder/releases/download/latest/hi3518ev100_lite_hiwatch-ds-i203-nor.tgz'
+
+# Board-specific U-Boot environment.
+fw_setenv osmem 128M
+fw_setenv mtdparts 'hi_sfc:256k(boot),64k(env),3072k(kernel),10240k(rootfs),-(rootfs_data)'
+fw_setenv phyaddru 3
+fw_setenv extras 'hieth.phyaddru=3 hieth.mdioifu=0'
```
openipc-ai, line 11: `extras` is assigned wholesale, not appended, and it is the single catch-all that `CONFIG_BOOTARGS` tacks onto the command line (`... mtdparts=${mtdparts} ${extras}`). Firmware's own generic customizer uses the same slot for something else entirely (`fw_setenv extras initcall_blacklist=hisi_femac_driver_init`).

Not a bug today, but anyone who later needs an initcall blacklist or an mmz tweak on this board has to remember to re-append `hieth.phyaddru=3 hieth.mdioifu=0` by hand, or Ethernet silently falls back to `CONFIG_HIETH_PHYID_U=1` from the kernel config. A one-line comment saying `extras` is assigned rather than appended would save that.

devices/hi3518ev100_lite_hiwatch-ds-i203/general/overlay/usr/share/openipc/customizer.sh
```diff
@@ -0,0 +1,28 @@
...
+fw_setenv sensor imx122
+
+cli -s .isp.sensorConfig /etc/sensors/imx122_spi_dc_1080p.ini
+
+cli -s .nightMode.irCutEnabled true
```
openipc-ai, line 16: `irCutEnabled` and `backlightEnabled` (line 20) are both already `true` by default — the wiki's `majestic-config.md` shows them commented out at `true`, and they are opt-out keys: only an explicit `false` parks the filter or the lamp. No other device in the tree sets either.

The pins on lines 17-18 and 21 are what actually configure this board, so these two lines can go. Keeping them mostly risks a future reader assuming they are load-bearing.

devices/hi3518ev100_lite_hiwatch-ds-i203/br-ext-chip-hisilicon/configs/hi3518ev100_lite_hiwatch-ds-i203_defconfig
```diff
@@ -0,0 +1,58 @@
...
+BR2_PACKAGE_HISILICON_OSDRV_HI3516CV100=y
+BR2_PACKAGE_IPCTOOL=y
+BR2_PACKAGE_LIBCURL_OPENIPC=y
+BR2_PACKAGE_LIBCURL_OPENIPC_CURL=y
+# BR2_PACKAGE_LIBCURL_OPENIPC_PROXY_SUPPORT is not set
+# BR2_PACKAGE_LIBCURL_OPENIPC_COOKIES_SUPPORT is not set
+# BR2_PACKAGE_LIBCURL_OPENIPC_EXTRA_PROTOCOLS_FEATURES is not set
+BR2_PACKAGE_LIBEVENT_OPENIPC=y
+BR2_PACKAGE_LIBOGG_OPENIPC=y
+BR2_PACKAGE_LINUX_FIRMWARE_OPENIPC=y
```
openipc-ai, line 51: This can go — the board has no WiFi and every sub-option was correctly dropped, but the parent is still enabled.

`linux-firmware-openipc.mk`'s `INSTALL_TARGET_CMDS` is four `ifeq` blocks keyed on `_ATHEROS_9271`, `_MEDIATEK_MT7601U`, `_RTL_8188EU` and `_XRADIOTECH_XR829`. With none of them selected the recipe expands to nothing, so the package is configured, built and tracked for zero bytes of output.

README.md
```diff
@@ -33,6 +33,7 @@ Foscam X5                SSC337DE     GC4653    RTL8188FU_USB    NOR_16M   done
 G.Craftsman GCA50        T31ZX        GC4653    -                NOR_16M   done
 EC37-T11                 T20L         SC2232    RTL8188FU_USB    MOR_16M   in progress
 H3C TC2101               SSC337       JXQ03     RTL8188FU_USB    NOR_16M   done
+HiWatch DS-I203          HI3518EV100  IMX122    -                NOR_16M   done, requires hi3518ev100-ddr3-256m U-Boot
```
openipc-ai, line 36: Sort order: `HiWatch` comes after `HB-WIFI-Z6 v1.2`, so this row wants to be one line further down. Column alignment is correct. (The table already has a couple of other misplacements, so this is drift rather than a new problem.)

openipc-ai (changes requested):
Thanks for the writeup — the hardware validation list and the build sizes made this a lot quicker to check, and the SoC model fix from the earlier round is right. That one mattered more than it looks: `rootfs_script.sh` resolves the exclude list as `${OPENIPC_SOC_MODEL}_${OPENIPC_VARIANT}.list`, so leaving `hi3516cv100` there would have orphaned the file.

Two things need changing before this can go in. Both are in `customizer.sh`, and both are cases where a camera brought up by hand behaves differently from a fresh install — which is probably why the end-to-end test passed:

1. `osmem 128M` is set without `totalmem`, and `load_hisilicon` hard-exits when `osmem >= totalmem`.
2. `fw_setenv mtdparts` rewrites the flash layout at first boot, three seconds before `reboot -f`.

Details inline. Everything after those two is a nit you can take or leave.

### Verified clean

- Defconfig delta against firmware's `hi3518ev100_lite_defconfig` is exactly the WiFi/mosquitto/motors removal plus `BR2_OPENIPC_FLASH_SIZE` 8 -> 16. Nothing else touched.
- Exclude list filename `hi3518ev100_lite.list` is correct, and the comment-only content is handled: `rootfs_script.sh` skips `#` lines, and firmware's own list for this board is empty anyway.
- Upgrade URL matches what CI publishes. The device name has two underscores, so `COMMON=2` takes the rename branch in `master.yml` and the asset lands as `hi3518ev100_lite_hiwatch-ds-i203-nor.tgz`.
- `${extras}` and `${mtdparts}` do reach the kernel command line — both appear in `CONFIG_BOOTARGS` in u-boot's `include/configs/hi-common.h`. `phyaddru` is read by u-boot's `drivers/net/hisfv300/net-drv.c`, and `hieth.phyaddru=3` works as a plain cmdline parameter because `CONFIG_HIETH_SWITCH_FABRIC=y` builds the driver in rather than as a module.
- `imx122` is a real branch in `hisilicon-osdrv-hi3516cv100`'s `load_hisilicon` (pinmux + `ssp_sony.ko`), and `imx122_spi_dc_1080p.ini` ships in that package and is installed to `/etc/sensors`.
- `.github/scripts/ci-matrix.py --self-test` passes on this branch, and the PR narrows the build matrix to exactly this one device.

PR #160, opened 2026-09-18
Add Imou Cue 2 (IPC-C22EN) device profile

HeytalePazguato:
## Summary

Add device profile for the **Imou Cue 2** (IPC-C22EN / IPC-C22EP), a sealed indoor camera built on Hi3516EV200 + SC2235 sensor over DVP parallel bus.

- Board defconfig based on `hi3516ev200_lite` with RTL8188FU WiFi and WPA supplicant
- Customizer sets U-Boot env vars: sensor, sensor_dvp, wlandev, gpio_button, upgrade URL
- LED init scripts (S01leds: red during boot, S99leds: green when ready)
- Exclude list strips unused sensor drivers/configs/IQ files to fit 8MB NOR flash
- Ships `wpa_passphrase` binary for WiFi configuration

Companion firmware PR: https://github.com/OpenIPC/firmware/pull/2446

## Hardware notes

The Imou Cue 2 runs **Dahua's signed U-Boot with RSA secure boot**. The UART console is locked and `loady`/`sf` commands are disabled. Flashing requires a **CH341A SPI programmer** connected directly to the NOR flash chip.

## Testing

Tested on **4 cameras** (IPC-C22EN). All boot cleanly from cold power with working 1080p video, WiFi, two-way audio (RTSP backchannel), day/night switching, and ONVIF discovery. Verified stable across multiple power cycles with no manual intervention required.

## Changes

- `devices/hi3516ev200_lite_imou-cue2-c22en/` — full device profile (defconfig, customizer, LED scripts, exclude list, wpa_passphrase)


devices/hi3516ev200_lite_imou-cue2-c22en/br-ext-chip-hisilicon/configs/hi3516ev200_lite_imou-cue2-c22en_defconfig
```diff
@@ -0,0 +1,65 @@
...
+BR2_PACKAGE_DROPBEAR_OPENIPC=y
+BR2_PACKAGE_HISILICON_OPENSDK=y
+BR2_PACKAGE_HISILICON_OSDRV_HI3516EV200=y
+BR2_PACKAGE_IPCTOOL=y
+BR2_PACKAGE_LIBCURL_OPENIPC=y
+BR2_PACKAGE_LIBCURL_OPENIPC_CURL=y
+# BR2_PACKAGE_LIBCURL_OPENIPC_PROXY_SUPPORT is not set
+# BR2_PACKAGE_LIBCURL_OPENIPC_COOKIES_SUPPORT is not set
+# BR2_PACKAGE_LIBCURL_OPENIPC_EXTRA_PROTOCOLS_FEATURES is not set
+# BR2_PACKAGE_LIBEVENT_OPENIPC is not set
```
openipc-ai, line 54: `majestic.mk` has a hard `MAJESTIC_DEPENDENCIES += libevent-openipc`, so Buildroot builds and installs libevent regardless of this symbol. The line saves no flash — it just makes the defconfig say something untrue. All 118 base defconfigs set `BR2_PACKAGE_LIBEVENT_OPENIPC=y`; please match them.

(The `WIREGUARD_*` and `VTUND` opt-outs above and the dropped `MAJESTIC_AF` / `MOTORS` are real savings and look right for this board.)

devices/hi3516ev200_lite_imou-cue2-c22en/br-ext-chip-hisilicon/configs/hi3516ev200_lite_imou-cue2-c22en_defconfig
```diff
@@ -0,0 +1,65 @@
...
+BR2_PACKAGE_LIBOGG_OPENIPC=y
+BR2_PACKAGE_MAJESTIC_FONTS=y
+BR2_PACKAGE_MAJESTIC_WEBUI=y
+BR2_PACKAGE_MAJESTIC=y
+BR2_PACKAGE_MBEDTLS_OPENIPC=y
+BR2_PACKAGE_OPUS_OPENIPC=y
+BR2_PACKAGE_OPUS_OPENIPC_FIXED_POINT=y
+BR2_PACKAGE_RTL8188FU_OPENIPC=y
+# BR2_PACKAGE_VTUND_OPENIPC is not set
+BR2_PACKAGE_WPA_SUPPLICANT=y
```
openipc-ai, line 64: This needs the other two symbols:

```
BR2_PACKAGE_WPA_SUPPLICANT=y
BR2_PACKAGE_WPA_SUPPLICANT_CLI=y
BR2_PACKAGE_WPA_SUPPLICANT_PASSPHRASE=y
```

Every other device in the tree sets all three. `_PASSPHRASE=y` is what installs the real `wpa_passphrase` binary that `/etc/network/interfaces.d/wlan0` calls — with it, the shim added in this PR can be deleted.

devices/hi3516ev200_lite_imou-cue2-c22en/general/overlay/usr/bin/wpa_passphrase
```diff
@@ -0,0 +1,8 @@
+#!/bin/sh
+cat <<CONF
+network={
+	ssid="$1"
+	#psk="$2"
+	psk="$2"
+}
+CONF
```
openipc-ai, line 8: Please drop this file and set `BR2_PACKAGE_WPA_SUPPLICANT_PASSPHRASE=y` in the defconfig instead — it is a workaround for a missing config symbol, not a device delta.

It is also strictly worse than the binary it stands in for. The real `wpa_passphrase` emits the PBKDF2-derived PMK as hex and keeps the cleartext only on the `#psk=` comment line, which `wlan0`'s `sed 's/#psk.*/scan_ssid=1/g'` then overwrites. This version emits `psk="$2"`, so the plaintext WiFi password ends up in `/tmp/wpa_supplicant.conf` — and a password containing a `"` or a backslash breaks the file outright. On top of that, it sits at `/usr/bin/wpa_passphrase`, the same path the real tool installs to, so it would silently shadow it the moment the option is turned on.

devices/hi3516ev200_lite_imou-cue2-c22en/general/overlay/usr/share/openipc/customizer.sh
```diff
@@ -0,0 +1,24 @@
+#!/bin/sh
+#
+# Perform basic settings on a known IP camera
+#
+#
+# Set SoC and sensor
+#
+fw_setenv soc hi3516ev200
+fw_setenv sensor sc2235
+fw_setenv sensor_dvp 1
```
openipc-ai, line 10: `sensor_dvp` is currently read only by `general/package/goke-osdrv-gk7205v200/files/script/load_goke`. Nothing on the HiSilicon side looks at it, so on firmware master this line is a no-op — it only starts working once firmware#2446's `load_hisilicon` change lands.

devices/hi3516ev200_lite_imou-cue2-c22en/general/overlay/usr/share/openipc/customizer.sh
```diff
@@ -0,0 +1,24 @@
...
+#
+# Set custom upgrade url
+#
+fw_setenv upgrade 'https://github.com/OpenIPC/builder/releases/download/latest/hi3516ev200_lite_imou-cue2-c22en-nor.tgz'
+#
+# Set reset button GPIO
+#
+fw_setenv gpio_button 56
```
openipc-ai, line 18: Nothing in the firmware rootfs reads `gpio_button`, and no other builder device sets it — is this consumed by the u-boot you flash, or is it left over from the stock environment? If it is dead, please drop it; if it is live, a one-line comment saying who reads it would help the next person.

devices/hi3516ev200_lite_imou-cue2-c22en/general/overlay/usr/share/openipc/customizer.sh
```diff
@@ -0,0 +1,24 @@
...
+#
+# Set wireless device and MAC address
+#
+fw_setenv wlandev rtl8188fu-hi3516ev200-imou-cue2
```
openipc-ai, line 22: This profile does not exist. `/etc/wireless/usb` is a chain of `if [ "$1" = ... ]` blocks ending in a bare `exit 1`, so an unknown name loads no module, prints nothing, and WiFi never appears — it looks like a driver fault rather than a missing stanza.

You do not need firmware#2446 for this. 23 builder devices ship their own `general/overlay/etc/wireless/usb`, which shadows the firmware copy for this board only. The sibling `devices/ssc325_lite_imou-c22cp` is the whole pattern:

```sh
#!/bin/sh

set_gpio() {
	[ "$2" -eq 1 ] && gpio set $1 || gpio clear $1
	sleep 1
}

if [ "$1" = "rtl8188fu-hi3516ev200-imou-cue2" ]; then
	set_gpio 52 1
	modprobe 8188fu
	exit 0
fi

exit 1
```

Adding that here (with whatever polarity GPIO 52 actually needs) drops one firmware dependency and keeps the board self-contained.

devices/hi3516ev200_lite_imou-cue2-c22en/general/overlay/usr/share/openipc/customizer.sh
```diff
@@ -0,0 +1,24 @@
...
+
+exit 0
```
openipc-ai, line 24: The customizer seeds no majestic settings at all. Right now those live in firmware#2446's `general/overlay/etc/majestic.yaml`, which is the global default for every board — the per-device values belong here as `cli -s` lines, the way `devices/ssc325_lite_imou-c22cp` and `devices/hi3516cv200_lite_spezvision-svi-252b-k202` do it. That covers the nightMode pins, codec/fps, the audio settings your PR description mentions, and `.isp.sensorConfig` if the SC2235 needs a non-default ini.

devices/hi3516ev200_lite_imou-cue2-c22en/general/scripts/excludes/hi3516ev200_lite.list
```diff
@@ -0,0 +1,72 @@
...
+/etc/sensors/sp2305_i2c_1080p.ini
+/etc/sensors/sp2308_i2c_1080p.ini
+#
+/etc/sensors/iq/f23.ini
+/etc/sensors/iq/gc2053.ini
+/etc/sensors/iq/imx307.ini
+/etc/sensors/iq/imx335.ini
+/etc/sensors/iq/sc2232.ini
+/etc/sensors/iq/sc2335.ini
+/etc/sensors/iq/default.ini
```
openipc-ai, line 60: Three of these seven were never installed on this SoC, and the other four leave the image with no IQ profile whatsoever.

`hisilicon-osdrv-hi3516ev200` installs exactly three IQ files — `imx307.ini`, `imx335.ini`, `f23.ini` — plus `default.ini` as a symlink to `imx307.ini`. So `gc2053.ini`, `sc2232.ini` and `sc2335.ini` (lines 55, 58, 59) match nothing and can go. More importantly, line 60 removes the fallback symlink the osdrv installs on purpose, and lines 54/56/57 remove every real profile, so `/etc/sensors/iq/` ends up empty unless firmware#2446's `sc2235.ini` is present. Same story one block up: every `/etc/sensors/*.ini` is stripped, so there is no sc2235 sensor config either.

Worth keeping `default.ini` unless you have measured that you need those few kB.

devices/hi3516ev200_lite_imou-cue2-c22en/general/scripts/excludes/hi3516ev200_lite.list
```diff
@@ -0,0 +1,72 @@
...
+#
+/etc/sensors/high-fps/imx307_1280x720_60fps.ini
+/etc/sensors/high-fps/imx307_1920x1080_60fps.ini
+/etc/sensors/high-fps/imx307_368x304_200fps.ini
+/etc/sensors/high-fps/imx307_640x480_130fps.ini
+/etc/sensors/high-fps/imx335_1280x720_60fps.ini
+/etc/sensors/high-fps/imx335_2592x1520_25fps.ini
```
openipc-ai, line 67: Neither filename exists. The hi3516ev200 osdrv ships `imx335_1280x720_120fps.ini`, `imx335_1296x972_64fps.ini`, `imx335_1920x1080_55fps.ini`, `imx335_2592x1944_45fps.ini` and `imx335_800x480_240fps.ini` — so these two entries prune nothing and all five real files stay in the image, which is the opposite of what the block intends. Substituting the five correct names buys back flash and drops two stale entries.

Altogether the list has five entries that match no file; `rootfs_script.sh` will print them as `excludes: <path> matched no file` with a `5 of 67` summary in the build log.

devices/hi3516ev200_lite_imou-cue2-c22en/general/overlay/etc/init.d/S01leds
```diff
@@ -0,0 +1,7 @@
+#!/bin/sh
+# Imou Cue 2: early boot — red LED on (GPIO 9 active low)
+case "$1" in
+  start)
+    gpio set 0; gpio clear 9; gpio clear 4
+    ;;
```
openipc-ai, line 6: The comment documents GPIO 9 but the line drives three pins (0, 4 and 9) — worth naming what 0 and 4 are, since nobody with the case open can guess. Also consider a `stop)` branch for symmetry with S99leds.

openipc-ai (changes requested):
Thanks for this — the profile is well shaped and four tested cameras is a good basis. The problem is that it is only half of the change, and two of the three dependencies on the firmware side do not need to exist. Requesting changes; details inline, plus the cross-cutting parts here.

## Blocking: nothing here works without firmware#2446, which is still open

That PR carries the SC2235 DVP register sequence, `sensor_dvp` support in `load_hisilicon`, the sc2235 sensor ini and IQ profile, and the `rtl8188fu-hi3516ev200-imou-cue2` stanza in `/etc/wireless/usb`. I ran this exclude list against what the build actually installs on firmware master. The complete set of sensor-related files that survive into the image is:

```
/usr/lib/sensors/libsns_sc2235.so
/etc/sensors/high-fps/imx335_1280x720_120fps.ini
/etc/sensors/high-fps/imx335_1296x972_64fps.ini
/etc/sensors/high-fps/imx335_1920x1080_55fps.ini
/etc/sensors/high-fps/imx335_2592x1944_45fps.ini
/etc/sensors/high-fps/imx335_800x480_240fps.ini
```

No sensor config, no IQ profile at all, and no WiFi. So this cannot merge before its firmware half — but two of those dependencies are avoidable today, see the comments on `customizer.sh` and `wpa_passphrase`.

## README row missing

Step 7 of CLAUDE.md: the device table in `README.md` needs a row. Worth clarifying the title's "IPC-C22EN / IPC-C22EP" too — `Imou IPC-C22EP-S2` is already in the table as an SSC325DE board, so if the C22EP really is this board it needs a clones-table row rather than a second meaning for the same model number.

## Worth reconsidering in firmware#2446

`etc/init.d/S01leds`, `etc/init.d/S99leds`, `etc/majestic.yaml`, `etc/rc.local`, `etc/ir/*` and `general/scripts/excludes/hi3516ev200_lite.list` all sit in firmware's *shared* overlay, so they would land on every OpenIPC board, not just this one — that is exactly the split builder exists to avoid, and all of them belong in this PR instead. What genuinely belongs in firmware is the opensdk SC2235 patch, the `load_hisilicon` change and the osdrv sensor config + IQ ini. Note also that replacing the SC2235 init table changes that driver for every board using the sensor, across hi3516cv200 and hi3516cv300 as well as this family, so it deserves its own testing note. `etc/ir/nrxset` is committed empty there. Finally, the two PRs disagree on the board name: `hi3516ev200_lite_imou-cue2-c22en` here versus `hi3516ev200_lite_imou-cue2` there.

## What already checks out

`ci-matrix.py --self-test` passes on the branch (116 devices, 15 smoke, 39 cases), so registration is automatic and nothing needs adding to the workflows. File modes and the device directory naming are right, and the exclude list correctly spares `libsns_sc2235.so`. I have not run a local build, so the 8 MB NOR fit is still unverified on my side.

PR #162, opened 2026-09-19
ssc377_lite: add TP-Link Tapo C120 device profile

nzzane:
Adds `devices/ssc377_lite_tp-link-tapo-c120` for the TP-Link Tapo C120: SSC377, SmartSens SC430AI (4 MP), 16 MB NOR (XM25QH128C), RTL8188FTV on USB.

What the device carries, and why each piece is here rather than in firmware:

- `ssc377_lite_tp-link-tapo-c120_defconfig` — the generic `ssc377_lite` defconfig with `BR2_OPENIPC_FLASH_SIZE="16"`, `BR2_PACKAGE_RTL8188FU_OPENIPC=y`, and `BR2_PACKAGE_WPA_SUPPLICANT_OPENIPC=y` in place of the stock supplicant (OpenIPC/firmware#2449; WPA3-SAE, needed for WPA3-only networks). Kernel config is the stock `infinity6c-ssc027a.config`, unchanged: `cfg80211` as a module is fine with `8188fu`.
- `general/overlay/etc/wireless/usb` — the `rtl8188fu-tapo-c120` entry: the Wi-Fi module's power is gated by GPIO 42, so it is driven high before `modprobe 8188fu`.
- `customizer.sh` — upgrade URL, `wlandev`, `sensor`, and majestic's IR-cut (GPIO 81, single wire, inverted), backlight (GPIO 14) and opus audio.
- `S96irled` — mirrors the IR-cut state onto the two IR LED GPIOs (12, 13); majestic has one backlight pin and this board has two LED banks.
- `sigmastar-osdrv-infinity6c/files/sensor/configs/sc430ai.bin` — the SC430AI IQ file, from the camera's own stock firmware. Same arrangement as `ssc335_lite_tp-link-tapo-c310-v220` (`sc3335.bin`, `sc3338.bin`); it was deliberately kept out of OpenIPC/firmware#2445 under that repo's binary policy. If it is not wanted here either, say so and I will drop it and keep the "take it from your own dump" instruction on the wiki.
- `excludes/ssc377_lite.list` — the other infinity6c sensor modules and IQ files.

Depends on OpenIPC/sensors#7 (driver, built by `sigmastar-osdrv-sensors` at HEAD), OpenIPC/firmware#2445 (`load_sigmastar` entry) and OpenIPC/firmware#2449 (the supplicant package). Until those merge, CI here cannot build this device, so this is a draft until they do.

**Built and tested** locally against those three branches, on the camera (flashed with `sysupgrade --kernel --rootfs`):

- `uImage.ssc377` 2039304 B (cap 2048 KB), `rootfs.squashfs.ssc377` 5513216 B (cap 8192 KB at 16 MB).
- `lsmod`: `cfg80211 184320 1 8188fu`; `wpa_cli status`: `key_mgmt=SAE pmf=2 wpa_state=COMPLETED` on a WPA3-only AP; DHCP.
- `ipcinfo -s` → `sc430ai`; majestic `Sensor index 0: 2688x1520@30fps`, `Load api bin Success - /etc/sensors/sc430ai.bin` (found by sensor name, no `sensorConfig` set); RTSP and JPEG fine, colour correct with IR-cut in the day position.
- `customizer.sh` ran on first boot: `upgrade`, `wlandev`, `sensor` set in the environment.
- `ci-matrix.py --self-test` passes (116 devices).

Known limitation, from OpenIPC/firmware#2449: the stock `wlan0` script still writes a WPA2-PSK block, so a WPA3-only network needs `key_mgmt=SAE` in the config by hand until the follow-up lands there.

Flash layout is the standard OpenIPC 16 MB map (`3072k(kernel),10240k(rootfs)`); the U-Boot image is the stock `u-boot-ssc377-nor.bin`. The wiki page (OpenIPC/wiki#549) is being updated to point at this device.


openipc-ai (approved):
All three dependencies are merged: OpenIPC/sensors#7 (0294a00), OpenIPC/firmware#2445 (e155691a) and OpenIPC/firmware#2449 (4d145e6c). Taking this out of draft so it can build.

**On `sc430ai.bin`: keep it.** It is the right thing here for the reason you gave — `ssc335_lite_tp-link-tapo-c310-v220` already ships `sc3335.bin` and `sc3338.bin` at the same path, and this repository's own notes describe `kc110-board-support` shipping a device-specific binary through `SITE_METHOD = local`. Per-device vendor assets are what `devices/` is for; the firmware tree's binary gate is about what reaches *every* camera, which is exactly why keeping it out of #2445 was correct. Both halves of that were right.

The rest reads well:

- The defconfig is a three-line delta from generic `ssc377_lite` — flash size, the driver, the supplicant — and nothing else, which is the shape a device profile should have.
- `excludes/ssc377_lite.list` resolves against `${OPENIPC_SOC_MODEL}_${OPENIPC_VARIANT}.list`, which is the name `rootfs_script.sh` looks for, and every one of its 16 entries matches a real file, so `test_excludes_report.sh` will stay quiet.
- `wireless/usb` calls `gpio` and `modprobe` only; `gpio` is a shell script in the shared overlay, so it is on the image. The `gpio set` idiom matches five other devices here.
- `S96irled` explains *why* it exists in its header — majestic has one backlight pin and this board has two LED banks — which is the comment style this tree wants.

Merging once CI is green.

PR #163, opened 2026-09-20
svi-252b: set night mode transition delay

ArthurKoba:
Follow-up to #161 and OpenIPC/majestic#325.

Tested the updated Majestic on the physical SpezVision SVI-252B K202
(HI3516CV200 + IMX323).

With `transitionDelayMs: 0`, the day -> night transition is now clean and
the previously reported purple/magenta flash is gone.

On night -> day, a very brief purple frame was still visible while the
bistable IR-cut filter was returning to the daylight position.

Setting:

```yaml
nightMode:
  transitionDelayMs: 150
```

removes the remaining flash. With 150 ms, both day -> night and night -> day
transitions are clean on this hardware.

This sets `transitionDelayMs` to 150 ms as the device-specific default for
the SpezVision SVI-252B K202.


ArthurKoba:
@openipc-ai Could you please take a quick look at this follow-up PR?

[Majestic #325](https://github.com/OpenIPC/majestic/issues/325) has now been hardware-tested on the SVI-252B and closed. The 150 ms transition delay fixes the remaining night -> day purple flash, and this PR only carries that confirmed value into the device profile.

Thanks.

openipc-ai:
Merged, thank you — and thank you for taking it back to the bench rather than
stopping at "the reported flash is gone."

That is the part worth calling out. The original report was about dusk, the fix
cleared dusk, and it would have been entirely reasonable to close the loop there.
Going on to check dawn, finding the frame that nobody had reported, and then
establishing that 150 ms is what it takes on this board is what turned a fix into
a finished device profile.

The delay stays device-specific on purpose, and your result is a good illustration
of why: what you measured is a bistable filter physically swinging shut, so the
number is a property of this board's mechanism rather than of the SoC, the sensor
or the firmware. Anyone reading this later and copying the 150 into another
profile should treat it as a starting point to verify at dawn on their own
hardware, not as a value that transfers.

This also retires the caveat #161 shipped with — that profile went in noting the
purple frame as a known remainder tracked upstream, and there is now nothing left
outstanding on the SVI-252B K202 day/night path. The only thing still open on the
board is the IR illuminator GPIO, which is why `backlightPin` remains unset; if
you ever pin that down, that would be a welcome follow-up.

The build is on master now and will be in tonight's nightly.

ArthurKoba:
@openipc-ai Small follow-up to the SVI-252B profile: this only corrects the remaining comment about backlightPin. Hardware confirms that the IR illuminator is switched autonomously by the daylight-sensor circuitry, so there is no SoC-controlled backlight GPIO to resolve or configure.

PR: https://github.com/OpenIPC/builder/pull/165

